PR #40, opened 2020-05-05
Allow user to specify cluster ID

jgsqware:
Signed-off-by: Julien Garcia Gonzalez <julien@giantswarm.io>

Toward https://github.com/giantswarm/giantswarm/issues/9271

### DIFF
--- a/cmd/template/cluster/flag.go
+++ b/cmd/template/cluster/flag.go
@@ -2,16 +2,20 @@ package cluster
 
 import (
 	"net"
+	"regexp"
+	"strconv"
 	"strings"
 
 	"github.com/giantswarm/microerror"
 	"github.com/spf13/cobra"
 
+	"github.com/giantswarm/kubectl-gs/internal/key"
 	"github.com/giantswarm/kubectl-gs/pkg/aws"
 	"github.com/giantswarm/kubectl-gs/pkg/gsrelease"
 )
 
 const (
+	flagClusterID               = "cluster-id"
 	flagDomain                  = "domain"
 	flagMasterAZ                = "master-az"
 	flagName                    = "name"
@@ -33,6 +37,7 @@ const (
 )
 
 type flag struct {
+	ClusterID               string
 	Domain                  string
 	MasterAZ                string
 	Name                    string
@@ -55,6 +60,7 @@ type flag struct {
 
 func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.Domain, flagDomain, "", "Installation base domain.")
+	cmd.Flags().StringVar(&f.ClusterID, flagClusterID, "", "User-defined cluster ID.")
 	cmd.Flags().StringVar(&f.MasterAZ, flagMasterAZ, "", "Tenant master availability zone.")
 	cmd.Flags().StringVar(&f.Name, flagName, "", "Tenant cluster name.")
 	cmd.Flags().BoolVar(&f.NoCache, flagNoCache, false, "Force updating release folder.")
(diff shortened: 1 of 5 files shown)


cmd/template/cluster/flag.go
```diff
@@ -78,6 +84,33 @@ func (f *flag) Init(cmd *cobra.Command) {
...
+			return microerror.Maskf(invalidFlagError, "--%s must be alphanumeric", flagClusterID)
+		}
+
+		matched, err := regexp.MatchString("^[a-z]+$", f.ClusterID)
+		if err == nil && matched == true {
+			// strings is letters only, which we also avoid
+			return microerror.Maskf(invalidFlagError, "--%s must be alphanumeric", flagClusterID)
+		}
+
+		if strings.ContainsAny(f.ClusterID, "1l") {
```
tfussell, line 103: Looks like this is covered by the case below. Is this just to have a nicer error message for these two characters in particular?

cmd/template/cluster/flag.go
```diff
@@ -78,6 +84,33 @@ func (f *flag) Init(cmd *cobra.Command) {
 func (f *flag) Validate() error {
 	var err error
 
+	if f.ClusterID != "" {
+		if len(f.ClusterID) != key.IDLength {
+			return microerror.Maskf(invalidFlagError, "--%s must be length of %d", flagClusterID, key.IDLength)
+		}
+
+		if _, err := strconv.Atoi(f.ClusterID); err == nil {
```
tfussell, line 92: I think it would be better to stick with a regex here because (I think--haven't tested) "01234" would throw an error due to the leading 0 and would therefore be accepted.

internal/key/key.go
```diff
@@ -15,15 +15,15 @@ import (
 const (
 	// IDChars represents the character set used to generate cluster IDs.
 	// (does not contain 1 and l, to avoid confusion)
-	idChars = "023456789abcdefghijkmnopqrstuvwxyz"
+	IdChars = "023456789abcdefghijkmnopqrstuvwxyz"
```
tfussell, line 18:
```suggestion
	IDChars = "023456789abcdefghijkmnopqrstuvwxyz"
```

cmd/template/cluster/flag.go
```diff
@@ -78,6 +84,33 @@ func (f *flag) Init(cmd *cobra.Command) {
...
+		if err == nil && matched == true {
+			// strings is letters only, which we also avoid
+			return microerror.Maskf(invalidFlagError, "--%s must be alphanumeric", flagClusterID)
+		}
+
+		if strings.ContainsAny(f.ClusterID, "1l") {
+			return microerror.Maskf(invalidFlagError, "--%s cannot contain 1 or l", flagClusterID)
+		}
+
+		matched, err = regexp.MatchString("^[023456789abcdefghijkmnopqrstuvwxyz]+$", f.ClusterID)
```
tfussell, line 107: It would be nice to build this regex from `key.IDChars` so they don't get out of sync.

cmd/template/cluster/flag.go
jgsqware: yes as I was going step by step, but you right it's overhead

internal/key/key.go
```diff
@@ -15,15 +15,15 @@ import (
 const (
 	// IDChars represents the character set used to generate cluster IDs.
 	// (does not contain 1 and l, to avoid confusion)
-	idChars = "023456789abcdefghijkmnopqrstuvwxyz"
+	IDChars = "023456789abcdefghijkmnopqrstuvwxyz"
```
MarcelMue, line 18: IMO we can include `1` if `l` is not included.

jgsqware: This can be discussed later. Let's keep it here as it is currently IMO

cmd/template/cluster/flag.go
```diff
@@ -78,6 +84,24 @@ func (f *flag) Init(cmd *cobra.Command) {
...
 	var err error
 
+	if f.ClusterID != "" {
+		if len(f.ClusterID) != key.IDLength {
+			return microerror.Maskf(invalidFlagError, "--%s must be length of %d", flagClusterID, key.IDLength)
+		}
+
+		matched, err := regexp.MatchString("^([a-z]+|[0-9]+)$", f.ClusterID)
+		if err == nil && matched == true {
+			// strings is letters only, which we also avoid
```
marians, line 94: This should be "strings is letters only or numbers only", which we avoid for usability reasons

PR #48, opened 2020-05-27
Add instance distribution

jgsqware:


cmd/template/cluster/flag.go
```diff
@@ -29,12 +29,14 @@ const (
...
-	flagNodesMin             = "nodex-min"
-	flagNumAvailabilityZones = "num-availability-zones"
+	flagAvailabilityZones                   = "availability-zones"
+	flagAWSInstanceType                     = "aws-instance-type"
+	flagNodepoolName                        = "nodepool-name"
+	flagNodesMax                            = "nodex-max"
+	flagNodesMin                            = "nodex-min"
+	flagNumAvailabilityZones                = "num-availability-zones"
+	flagOnDemandBaseCapacity                = "on-demand-base-capacity"
+	flagOnDemandPercentageAboveBaseCapacity = "on-demand-precentage-above-base-capacity"
```
kopiczko, line 39: I'm not sure if those flags are descriptive enough. I'd like to have something the CR structure better. But my practical approach may lead to really ugly UX. Summoning @giantswarm/sig-ux here as they may have some good ideas.

corest: it is probably time to move to config files from flags

jgsqware: Do you want to wait for sig ux for this one, or create a main issue where we can tackle this properly, as Adidas is wainting for me to release the other change too

marians:
@jgsqware What's the goal of this PR?

jgsqware:
Let the customer choosing the base capacity and percentage spot instance of the instance distribution.

Currently they do it manually in the yaml after generation.

I an gladly take the refactoring in a config file if needed later.

marians:
So basically you are trying to modify node pool settings via command line flags that are made for cluster creation? Since the cluster to node pool relationship is one : many I don't like that approach a lot. This basically limits the command to creation of only a single node pool.

jgsqware:
@marians on the current way of doing it, you can only create 1 node pools or 0, then you can generae CR to add new one.

cmd/template/cluster/flag.go
kopiczko: I don't want to make the decision here as I'm not involved enough in development of the plugin. I'm just alerted.

paurosello:
I think we need to decide the intended way of using kubectl-gs:
- Use kubectl-gs to make the skeleton of the cluster, user modifies and adds whatever it needs and stores the yamls in a git repo.
- Use kubectl-gs to continuously generate the cluster with all parameters and send that directly to the kubernetes API.

From my POV this tools was intended for the first case, as you can only create a single nodepool now for example.
Terraform might be a better tool to achieve the second case.

I thing this parameters should be added to the nodepool creation and not to the cluster creation. The default nodepool created with the cluster should always contain the default values.

marians:
I suggest we extract the node pool manifest creation into a separate command.

marians:
Ignore my comment above, as it was based on a false assumption.

marians:
FYI: I'd like to drop the default node pool creation, as it complicates things. See https://github.com/giantswarm/kubectl-gs/pull/49

jgsqware:
I will rebase.

jgsqware:
Rebased and removed from clusterCR. Merging

corest:
And @paurosello is right - this tools was written to create skeleton, not to generate files continuously

PR #49, opened 2020-05-28
Remove default node pool creation from `kubectl gs template cluster`

marians:
This removes the default node pool creation from the cluster templating.

Reason: With this function, there are two ways for users to think about node pools. One for the first node pool, the other for all other node pools. This actually complicates things for users. Plus it bloats the code base.

In `gsctl` we introduced the "default node pool" concept when we added node pool functionality for AWS, so that it would still be easy and simple to create a functional cluster by using a relatively short command. We basically had existing users with adopted behaviour in mind. This is not relevant for kubectl-gs.

So for user, with this change it will be:

1. Run `kubectl gs template cluster` once
2. Run `kubectl gs template nodepool` as often as required

### DIFF
--- a/cmd/template/cluster/flag.go
+++ b/cmd/template/cluster/flag.go
@@ -3,7 +3,6 @@ package cluster
 import (
 	"net"
 	"regexp"
-	"strings"
 
 	"github.com/giantswarm/microerror"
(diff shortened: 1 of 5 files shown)


corest:
Didn't like either. +1

PR #53, opened 2020-06-09
Add credential setting and explain how to get it

pipo02mix:


cmd/template/cluster/flag.go
```diff
@@ -44,6 +46,7 @@ type flag struct {
 func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.Domain, flagDomain, "", "Installation base domain.")
 	cmd.Flags().StringVar(&f.ClusterID, flagClusterID, "", "User-defined cluster ID.")
+	cmd.Flags().StringVar(&f.Credential, flagCredential, "credential-default", "Cloud provider credentials used to spin up the cluster.")
```
pipo02mix, line 49: fallbacks to default credential

docs/template-cluster-cr.md
```diff
@@ -18,6 +18,7 @@ It supports the following flags:
   Use the flag once with a single value to create a cluster with one master node. For master node high availability,
   specify three distinct availability zones instead. This can be done by separating AZ names with comma or using the flag
   three times with a single AZ name.
+- `--credential` - AWS cloud credentials that point to the AWS account used to spin up the cluster resources. To get this info run against the Control Plane API `kubectl -n giantswarm get secret -oyaml | grep ORG_NAME -A2 | tail -n 1 | awk '{print $2}'` replacing `ORG_NAME` for the name of the organization selected.
```
pipo02mix, line 21: We dont want to add gs API logic here. It is just a templating tool so users need to provide the parameters. Ideally in official plugin this is deducted from org name (same as base domain or others).

docs/template-nodepool-cr.md
```diff
@@ -13,16 +13,16 @@ To create the manifests for a new node pool, use this command:
 
 Here are the supported flags:
 
-  - `--availability-zones` - list of availability zones to use, instead of setting a number. Use comma to separate values.
+  - `--availability-zones` - list of availability zones to use, instead of setting a number. Use comma to separate values. (e.g. `eu-central-1a,eu-central-1b`)
```
omegas27, line 16: Should we not call it failure domain  to be closer to the cluster api?

omegas27:
Nice work. 
This is not on you but I fear we are trying to make it way too close to the AWS implementation and I think this will lead to issues later when we want to add Azure here instead of trying to find a common one like the cluster api

docs/template-nodepool-cr.md
pipo02mix: well we align with k8s terminology but IMO we move away for end-user, we have been naming this `availability zones` since the beginning. If we do a consensual change cross-company we can change terminology in all places (UI, ctls, docs) as we did with BYOC. But I will not start here

pipo02mix:
Yes you are right this getting close to AWS. But this is temporal plugin born in a hackathon as tool for templating, it lacks of lot of functionality. We will have an [official and support plugin](https://github.com/giantswarm/kubectl-gigantic) that gives more functionality and does more than template (create cluster indeed, validation,...) So this is just temporal helper for customer started with beta control plane API

pipo02mix:
I go ahead @omegas27

omegas27:
It works for me. I just wanted to raise my concerns as I am still not sure how I should mention them :)
Thanks for your answers

PR #55, opened 2020-06-24
Allow specifying tenant cluster labels through --label flags

ubergesundheit:
Towards giantswarm/giantswarm#11403

adds a `--label` flag to the `template cluster` command, which allows to set tenant cluster labels added to the `Cluster` CR.

### DIFF
--- a/cmd/template/cluster/flag.go
+++ b/cmd/template/cluster/flag.go
@@ -1,6 +1,7 @@
 package cluster
 
 import (
+	"github.com/giantswarm/kubectl-gs/pkg/clusterlabels"
 	"net"
 	"regexp"
 
@@ -26,6 +27,7 @@ const (
 	flagOwner        = "owner"
 	flagRegion       = "region"
 	flagRelease      = "release"
+	flagLabel        = "label"
 )
 
 type flag struct {
@@ -41,6 +43,7 @@ type flag struct {
 	Owner        string
 	Region       string
 	Release      string
+	Label        []string
 }
 
 func (f *flag) Init(cmd *cobra.Command) {
@@ -56,6 +59,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.Owner, flagOwner, "", "Tenant cluster owner organization.")
 	cmd.Flags().StringVar(&f.Region, flagRegion, "", "Installation region (e.g. eu-central-1).")
 	cmd.Flags().StringVar(&f.Release, flagRelease, "", "Tenant cluster release.")
+	cmd.Flags().StringSliceVar(&f.Label, flagLabel, []string{}, "Tenant cluster label.")
 }
 
 func (f *flag) Validate() error {
@@ -134,6 +138,11 @@ func (f *flag) Validate() error {
 
 	if !release.Validate(f.Release) {
 		return microerror.Maskf(invalidFlagError, "--%s must be a valid release", flagRelease)
+	}
+
+	_, err = clusterlabels.Parse(f.Label)
+	if err != nil {
+		return microerror.Maskf(invalidFlagError, "--%s must contain valid label definitions (%s)", flagLabel, err)
 	}
 
 	return nil
--- a/cmd/template/cluster/flag.go
+++ b/cmd/template/cluster/flag.go
@@ -59,7 +59,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.Owner, flagOwner, "", "Tenant cluster owner organization.")
 	cmd.Flags().StringVar(&f.Region, flagRegion, "", "Installation region (e.g. eu-central-1).")
 	cmd.Flags().StringVar(&f.Release, flagRelease, "", "Tenant cluster release.")
-	cmd.Flags().StringSliceVar(&f.Label, flagLabel, []string{}, "Tenant cluster label.")
+	cmd.Flags().StringSliceVar(&f.Label, flagLabel, nil, "Tenant cluster label.")
 }
 
 func (f *flag) Validate() error {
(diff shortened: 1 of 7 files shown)


cmd/template/cluster/flag.go
```diff
@@ -56,6 +59,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.Owner, flagOwner, "", "Tenant cluster owner organization.")
 	cmd.Flags().StringVar(&f.Region, flagRegion, "", "Installation region (e.g. eu-central-1).")
 	cmd.Flags().StringVar(&f.Release, flagRelease, "", "Tenant cluster release.")
+	cmd.Flags().StringSliceVar(&f.Label, flagLabel, []string{}, "Tenant cluster label.")
```
axbarsan, line 62:
```suggestion
	cmd.Flags().StringSliceVar(&f.Label, flagLabel, nil, "Tenant cluster label.")
```

docs/template-cluster-cr.md
```diff
@@ -27,6 +27,7 @@ It supports the following flags:
 - `--region` - tenant cluster AWS region. Must be configured with installation region.
 - `--release` - valid release version.
   Can be retrieved with `gsctl list releases` for your installation. Only versions above *10.x.x*+ support cluster CRs.
+- `--label` - tenant cluster label in form of `key=value`. Can be specified multiple times. Only clusters with release version above *10.x.x*+ support tenant cluster labels.
```
axbarsan, line 30:
```suggestion
- `--label` - tenant cluster label in the form of `key=value`. Can be specified multiple times. Only clusters with release version above *10.x.x*+ support tenant cluster labels.
```

pkg/clusterlabels/labels.go
```diff
@@ -0,0 +1,36 @@
...
+			if errs := validation.IsQualifiedName(parts[0]); len(errs) != 0 {
+				return nil, fmt.Errorf("invalid label key: %q: %s", labelSpec, strings.Join(errs, ";"))
+			}
+			if errs := validation.IsValidLabelValue(parts[1]); len(errs) != 0 {
+				return nil, fmt.Errorf("invalid label value: %q: %s", labelSpec, strings.Join(errs, ";"))
+			}
+			labels[parts[0]] = parts[1]
+		} else {
+			return nil, fmt.Errorf("unknown label spec: %v", labelSpec)
+		}
```
axbarsan, line 33: Is checking if the label contains the `=` token really necessary? We already check if the label format works without it.

```suggestion		
		parts := strings.Split(labelSpec, "=")
		if len(parts) != 2 {
			return nil, fmt.Errorf("invalid label spec: %v", labelSpec)
		}
		if strings.Contains(parts[0], label.ForbiddenLabelKeyPart) {
			return nil, fmt.Errorf("invalid label key: %q: containing forbidden substring '%s'", labelSpec, label.ForbiddenLabelKeyPart)
		}
		if errs := validation.IsQualifiedName(parts[0]); len(errs) != 0 {
			return nil, fmt.Errorf("invalid label key: %q: %s", labelSpec, strings.Join(errs, ";"))
		}
		if errs := validation.IsValidLabelValue(parts[1]); len(errs) != 0 {
			return nil, fmt.Errorf("invalid label value: %q: %s", labelSpec, strings.Join(errs, ";"))
		}
		labels[parts[0]] = parts[1]
```

axbarsan: We would lose stack traces here by using `fmt.Errorf`, instead of `microerror.Maskf`

pkg/template/cluster/cluster.go
```diff
@@ -68,6 +69,23 @@ func newClusterCR(obj interface{}, clusterID string, c Config) (*apiv1alpha2.Clu
...
+
+	for key, value := range c.Labels {
+		clusterLabels[key] = value
+	}
+
+	gsLabels := map[string]string{
+		label.ClusterOperatorVersion: c.ReleaseComponents["cluster-operator"],
+		label.Cluster:                clusterID,
+		label.Organization:           c.Owner,
+		label.ReleaseVersion:         c.ReleaseVersion,
```
axbarsan, line 82: Is there any reason for not defining these as the initial values inside the `clusterLabels` map?

pkg/clusterlabels/labels.go
ubergesundheit: I used `microerror.Mask(fmt.Errorf(...))` now because the errors are created here.

axbarsan: We usually define errors in a separate file, in a format such as:

```go
var invalidLabelKeyError = &microerror.Error{
	Kind: "invalidLabelKeyError",
}

func IsInvalidLabelKey(err error) bool {
	return microerror.Cause(err) == invalidLabelKeyError
}
```

Which you would then call as:

```go
microerror.Maskf(invalidLabelKeyError, %q: containing forbidden substring '%s'", labelSpec, label.ForbiddenLabelKeyPart)
```

cmd/template/cluster/flag.go
axbarsan: `nil` is actually a slice's zero value (the value it has by default). A `nil` slice has a `nil` backing array. This means less computation. Also, the Go compiler applies several optimizations for `nil` slices, whereas empty slices can only be optimized at run time.

A more visual example (how a slice data structure looks in go)

* **nil slice**
```go
{
	capacity: 0
	length: 0
	backingArray: nil
}
```

* **empty slice**
```go
{
	capacity: 0
	length: 0
	backingArray: []
}
```

The actual performance difference won't be noticeable in this case, but it's good practice to use `nil` slices in general.

All built-in slice functions accept `nil` slices. That's actually the reason why we get a slice's length by using `len(slice)`, instead of `slice.length`. The only issue is when you marshal to JSON (a `nil` slice is serialized as `null`, as opposed to an empty array)

ubergesundheit: :pray: many thanks thats a good explanation!

pkg/template/cluster/cluster.go
ubergesundheit: The idea was that (although possibly prevented through the check for forbidden label keys), to always have the required labels overwrite userlabels.

PR #56, opened 2020-06-24
Update main README, Installation docs for Krew

marians:
This should beautify the repo README a bit.

WARNING: this assumes that we already have distribution via `krew` in place.


marians:
Jep, that's the plan

PR #58, opened 2020-07-01
Add packaging setup

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/11783

This PR adds:
* Makefile scripts for building, packaging, testing, installing, linting, etc.
* GitHub actions for automatic releases
* GitHub action for pushing latest version to `krew-index`


.gitignore
```diff
@@ -1 +1,2 @@
-kubectl-gs
+kubectl-gs*
```
marians, line 1: What does the star mean here?

axbarsan: It means that it will ignore all files that begin with the `kubectl-gs` prefix.

When running, `make build-darwin`, you can actually have multiple versions of the executable side by side (in the format of `kubectl-gs-{TAG}-{OS}-{ARCH}`), so this helps us not include all of those in the repo

PR #64, opened 2020-07-02
Remove krew github action and add circleci config

axbarsan:
Apparently GitHub actions cannot be triggered by other GitHub actions, so we can't trigger the krew update action when a release is published (or a tag is created).

A workaround is to use CircleCI, which will always work. This workflow will fetch the binary that handles updating the krew-index, and run it, creating the PR in the krew-index.


kopiczko:
I don't get it. You can setup an action triggered only when certain tag patterns are pushed.

axbarsan:
I'll try that, I only tried by binding to release-specific events

.github/workflows/krew.yaml
```diff
@@ -1,8 +1,9 @@
 name: Push to Krew
 
 on:
-  release:
-    types: [published]
+  push:
+    tags:
+      - '*'
```
kopiczko, line 6: We can make this a bit stronger:

```suggestion
      - 'v*.*.*'
```

axbarsan:
> LGTM when works

Unfortunately it doesn't... I've tested a similar setup on a testing repo...

So I guess... back to CircleCI for updating krew-index? 😬

PR #65, opened 2020-07-02
Use 'kubectl gs' as executable name

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/11820

Not pretty. `Cobra` splits the root command's `Use` field by spaces, into multiple words, and only displays the first word.

With this, we're fooling it by using the hex value of a non-breakable space (NBSP char), instead of a regular space character.

## Preview

image

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -12,6 +12,11 @@ import (
 )
 
 const (
+	// Hack to set base command name as 'kubectl gs', since
+	// cobra splits all the words in the 'usage' field and
+	// only prints the first word. The splitting is done by
+	// space characters (' '), and we trick it by using a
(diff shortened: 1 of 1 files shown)


cmd/root.go
```diff
@@ -4,14 +4,15 @@ import (
...
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 	"github.com/spf13/cobra"
+
+	"github.com/giantswarm/kubectl-gs/cmd/template"
 )
 
 const (
-	name        = "kubectl-gs"
+	name        = "kubectl" + string(0xA0) + "gs"
```
ubergesundheit, line 15: Maybe add the explanation for using the non breakable space also here in the code as a comment?

PR #85, opened 2020-07-03
Add basic login command

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/11793

This is copy/paste from the same command, which was in the `kubectl-gigantic` repo.

Adding it as it is first, so I could then make changes to it and not have a huge Pull Request.


marians:
FYI: I just tested after approving, but could not manage to enter a valid URL. Every URL I tried resulted in 

```
Error: invalid flag error: --k8s-api-url flag must be a valid URL
panic: "{\"kind\":\"invalidFlagError\",\"annotation\":\"--k8s-api-url flag must be a valid URL\",\"stack\":[{\"file\":\"/Users/marian/git/giantswarm/kubectl-gs/cmd/login/flag.go\",\"line\":28},{\"file\":\"/Users/marian/git/giantswarm/kubectl-gs/cmd/login/runner.go\",\"line\":56},{\"file\":\"/Users/marian/git/giantswarm/kubectl-gs/main.go\",\"line\":57}]}
```

axbarsan:
That's strange. It works for me for both:
- https://g8s.gauss.eu-central-1.aws.gigantic.io
- https://g8s.ginger.eu-west-1.aws.gigantic.io

cmd/login/template/template.go
```diff
@@ -0,0 +1,28 @@
+package template
+
+import (
+	"io"
+
+	"github.com/giantswarm/microerror"
+	"github.com/markbates/pkger"
```
kopiczko, line 7: Just saying that in apiextension we use github.com/mjibson/esc now.

cmd/login/runner.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/giantswarm/kubectl-gs/cmd/login/template"
+	"github.com/giantswarm/kubectl-gs/pkg/callbackserver"
+	"github.com/giantswarm/kubectl-gs/pkg/installation"
+	"github.com/giantswarm/kubectl-gs/pkg/kubeconfig"
+	"github.com/giantswarm/kubectl-gs/pkg/oidc"
+)
+
+const (
+	clientID     = "zQiFLUnrTFQwrybYzeY53hWWfhOKWRAU"
+	clientSecret = "TVHzVPin2WTiCma6bqp5hdxgKbWZKRbz" // #nosec G101
```
pipo02mix, line 29: What are these values?

pipo02mix: I meant how they are generated? are static? can we expose them in code?
I imagine they are sued for oidc client, but I dont know exactly how it works

axbarsan: Yes, these are for OIDC. They are static and I've generated them using some random online tool.
We're using a dex [public client](https://github.com/dexidp/dex/blob/master/Documentation/custom-scopes-claims-clients.md#public-clients), so exposing these credentials is not a security issue

PR #87, opened 2020-07-08
Make errors more user-friendly

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/11983

This PR:
* Fixes the error output of unknown commands
* Adds an utility for pretty-printing errors (no more ugly stack traces):
  * It adds color output to the error title
  * It adds the `Error: ` prefix if the error title doesn't already have it
  * It capitalizes the first word in the error title
  * If using a microerror, and there is no custom message specified, the error kind is adapted, by removing the `error` suffix out of it

See more below:

## Preview

<details><summary>Microerror, with no additional message</summary>

```go
&microerror.Error{
	Kind: "invalidConfigError",
}
```

image

</details>

<details><summary>Microerror, with additional message</summary>

```go
err := &microerror.Error{
	Kind: "invalidConfigError",
}
microerror.Maskf(err, "something went terribly wrong")
```

image

</details>

<details><summary>Microerror, with additional multi-line message</summary>

```go
err := &microerror.Error{
	Kind: "invalidConfigError",
}
microerror.Maskf(err, "something went terribly wrong!\nThis almost never happens, though")
```

image

</details>


cmd/root.go
```diff
@@ -95,11 +95,12 @@ func New(config Config) (*cobra.Command, error) {
...
-		Short:        description,
-		Long:         description,
-		RunE:         r.Run,
-		SilenceUsage: true,
+		Use:           name,
+		Short:         description,
+		Long:          description,
+		RunE:          r.Run,
+		SilenceUsage:  true,
+		SilenceErrors: true,
```
axbarsan, line 103: This makes cobra shut up and not print errors by itself. Only we control how we print them.

cmd/root.go
```diff
@@ -95,11 +96,18 @@ func New(config Config) (*cobra.Command, error) {
...
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) > 0 {
+				return fmt.Errorf("unknown command %q for %s", args[0], cmd.CommandPath())
+			}
+			return nil
+		},
```
axbarsan, line 110: This makes the output of an unknown command (e. g. `kgs version` have the correct command in the output -> `kubectl gs` instead of `kubectl\u00a0gs`

pkg/errorprinter/errorprinter.go
```diff
@@ -0,0 +1,83 @@
...
+		for _, row := range errorRows[1:] {
+			messageBuilder.WriteString(row)
+			messageBuilder.WriteString("\n")
+		}
+	}
+
+	return messageBuilder.String()
+}
+
+func (ep *ErrorPrinter) formatTitle(title string) string {
```
kopiczko, line 50: They way we write error messages will make it very hard to debug if the thing before `:` and stack is not printed. I fully agree for some kinds we shouldn't print stack. Like for `invalidFlagError` which I'm trying to do [here](https://github.com/giantswarm/devctl/pull/67/files#diff-7c5bc5c38f66a3679f1d4ad81c524841R41-R43).

I'm not blocking this because I understand the output is really ugly for the end user but I also think debugging may become hard.

I think it might be an idea to invest some time into something like `microerror.Pretty` which would also print stacks in a digestable way + handling `invalidFlagError` separately.

We can also agree on some `userError` kind which would be printed the same way `invalidFlagError` is printed.

marians:
I'm all for it, of course! :)

To address @kopiczko 's concerns regarding debugging, maybe we should add a global `--debug` flag which enables the printing of a stack trace. Different PR though.

pkg/errorprinter/errorprinter.go
axbarsan: Adding some extra metadata to each microerror is definitely something I'll look into. Also, some built-in error types (like the `invalidFlagError`).

For now, I'll go with having a `--debug` flag (a separate PR will come) for viewing full error output

PR #88, opened 2020-07-13
kgs docs fixes and improvements

marians:
Fixed a few things, mainly regarding installation


docs/installation.md
```diff
@@ -26,41 +25,67 @@ alias kgs="kubectl gs"
...
+## Without Krew
 
 The platform-agnostic description:
 
 1. Download the [latest release](https://github.com/giantswarm/kubectl-gs/releases/latest) archive for your platform
 2. Unpack the archive
 3. Copy the executable to a location included in your `$PATH`
 4. Create an alias `kgs` for `kubectl gs`
-5. Check it's working by executing `kgs info`
+5. Check it's working by executing `kgs`
```
axbarsan, line 47:
```suggestion
5. Check if it's working by executing `kgs`
```

README.md
```diff
@@ -5,21 +5,23 @@
...
 ## Features
 
-- **Custom resource templating**: create manifests for creating/updating custom resources for:
+- **Custom resource templating**: using the `template` command lets you create manifests for
+  creating/updating custom resources for:
   - Clusters
   - Node pools
   - App catalogs
   - Apps
+- **SSO login**: with the `login` command you can quickly set up a `kubectl context` with
```
axbarsan, line 23: Do we want to include the `login` command so early? Not all installations support it

marians: Definitely. The command is there, so we should talk about it. The problem that not all installations support it is something we must take care of separately.

docs/installation.md
```diff
@@ -26,41 +25,67 @@ alias kgs="kubectl gs"
 Lastly, let's check that the plugin is working as it's supposed to.
```
pipo02mix, line 25:
```suggestion
Lastly, let's check that the plug-in is working as it's supposed to.
```

PR #98, opened 2020-07-13
Add flag for toggling debug mode

axbarsan:
Adds a new persistent `--debug` flag. This flag can be used for enabling full error output.

**Debug mode off:**

image

**Debug mode on:**

image

### DIFF
--- a/cmd/flag.go
+++ b/cmd/flag.go
@@ -4,10 +4,16 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	flagDebugMode = "debug"
+)
+
 type flag struct {
+	debugMode *bool
 }
 
 func (f *flag) Init(cmd *cobra.Command) {
+	cmd.PersistentFlags().BoolVar(f.debugMode, flagDebugMode, false, "Toggle debug mode, for seeing full error output.")
 }
 
 func (f *flag) Validate() error {
--- a/cmd/flag.go
+++ b/cmd/flag.go
@@ -5,16 +5,15 @@ import (
 )
 
 const (
-	flagDebugMode = "debug"
+	flagDebug = "debug"
 )
 
 type flag struct {
-	debugMode bool
 }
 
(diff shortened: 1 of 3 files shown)


main.go
```diff
@@ -15,12 +15,20 @@ import (
...
+	DebugMode = false
+)
+
 func main() {
 	err := mainE(context.Background())
 	if err != nil {
-		ep := errorprinter.New()
-		fmt.Print(ep.Format(err))
-		os.Exit(1)
+		if DebugMode {
```
kopiczko, line 25: Passing this pointer up is really ugly. How about doing `if containsString(args, "--debug")` and adding a flag only for `-h` output with a comment in code that it's ignored and handled in `main.go`?

cmd/flag.go
```diff
@@ -4,10 +4,17 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	flagDebugMode = "debug"
+)
+
 type flag struct {
+	debugMode bool
```
kopiczko, line 12:
```suggestion
```

cmd/flag.go
```diff
@@ -4,10 +4,17 @@ import (
...
 }
 
 func (f *flag) Init(cmd *cobra.Command) {
+	// This value is ignored. The real value is handled inside 'main.go'.
+	cmd.PersistentFlags().BoolVar(&f.debugMode, flagDebugMode, false, "Toggle debug mode, for seeing full error output.")
```
kopiczko, line 17:
```suggestion
	cmd.PersistentFlags().Bool(flagDebug, false, "Toggle debug mode, for seeing full error output.")
```

cmd/flag.go
```diff
@@ -4,10 +4,17 @@ import (
 	"github.com/spf13/cobra"
 )
 
+const (
+	flagDebugMode = "debug"
```
kopiczko, line 8:
```suggestion
	flagDebug = "debug"
```

PR #100, opened 2020-07-14
Add 'get clusters' command

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/11797

> ### ⚠️ This PR is up for **UX** review. It is also huge.
> Once approved, I will split it into multiple ones and ask for a code review on each one of them.

## In this PR

- Getting all clusters (v5 only) in a AWS or Azure installation
- Getting clusters by ID
- `TABLE`/`JSON`/`YAML`/`NAME`/`JSONPATH` output for getting clusters (by ID or not)

## Preview

<details><summary>Get clusters</summary>

image

</details>

<details><summary>Get cluster by ID</summary>

image

</details>

<details><summary>Get clusters, no clusters available</summary>

image

</details>

<details><summary>Get clusters by ID, no cluster found</summary>

image

</details>

<details><summary>Help</summary>

image

</details>


marians:
Going for a test ride 🚴 ...

marians:
`kgs get clusters` just works! Cool!

---

```
$ kgs get clusters --help
...
-o, --output string   Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file
...

$ kgs get clusters -o name
Error: *cluster.CommonClusterList
```

---

This produces no output:

```
$ kgs get clusters -o=jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}'
```

Example from https://kubernetes.io/docs/reference/kubectl/jsonpath/

---

It seems like the `-n`/ `--namespace` flag does nothing. All clusters get listed.

axbarsan:
I've added support for `--output name` and for using a custom `namespace`.

Jsonpath works, but with a drawback:
Since V4 clusters have to be wrapped in a custom list type (check the PR description), the jsonpath structure for V4 clusters would be a bit different:
`kgs get clusters -o=jsonpath='{range .items[*].items[*]}{.metadata.name}{"\n"}{end}'`

For V5 clusters the example you provided works: 
`kgs get clusters -o=jsonpath='{range .items[*]}{.metadata.name}{"\n"}{end}'`

marians:
To be honest, I am not sure it's worth including v4 with this command. I can't remember that we discussed this, however v4 clusters won't support full management via the CP k8s API anyway. Should we drop them? Sorry for the extra effort you had!

axbarsan:
💀 💀 💀 

Yeah, let's remove them, they only make stuff harder to handle

marians:
Do we have any influence on the date format?

image

I'd love to get rid of the `+0000`, as it is redundant with `UTC`.

axbarsan:
> Do we have any influence on the date format?
> I'd love to get rid of the `+0000`, as it is redundant with `UTC`.

Unfortunately we can't customize it. The table printer doesn't provide any customizing options for that.
The only thing it accepts is the `date-time` format:
https://github.com/giantswarm/kubectl-gs/blob/c5efc3865e7f7c5afd86e0c1a58cdcf9f320b201/cmd/get/clusters/provider/aws.go#L17

axbarsan:
I removed everything V4 related.

So currently, you can get V5 AWS and Azure clusters

marians:
Works well on AWS.

On an Azure installation (gollum, godsmack) I get this:

```
kgs get clusters
Error: User "marian@giantswarm.io" cannot list resource "clusters" in API group "cluster.x-k8s.io" in the namespace "default"
```

which seems to be a permission problem. It works when I use a key pair context instead of OIDC. I also checked that the CRD `clusters.cluster.x-k8s.io` is in place.

On ghost it is again slightly different:

```
kgs get clusters
Error: Unauthorized
```

axbarsan:
> ```
> kgs get clusters
> Error: User "marian@giantswarm.io" cannot list resource "clusters" in API group "cluster.x-k8s.io" in the namespace "default"
> ```

That is indeed a permission issue, I'll try to get to the bottom of it.

> On ghost it is again slightly different:
> 
> ```
> kgs get clusters
> Error: Unauthorized
> ```

I believe this is a backend issue on our side (with `dex` authentication), which prevents being able to renew expired tokens. Logging in again should work, for now. I'll ask about it on the WG channel

axbarsan:
But since both of those are back-end issues, I think I can split this up for code review, if you don't have anything against it

axbarsan:
Fix for the permissions issue
https://github.com/giantswarm/installations/pull/1372

axbarsan:
Closing because all of the other PRs have been merged 🥳

PR #104, opened 2020-07-22
Update architect orb/makefile/gh workflows latest versions

axbarsan:
## Checklist

- [ ] Update changelog in CHANGELOG.md. (Not important to the customer)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,38 +13,47 @@ SOURCES        := $(shell find . -name '*.go')
 VERSION        := $(shell architect project version)
 LDFLAGS        ?= -w -linkmode 'auto' -extldflags '-static' \
   -X '$(shell go list .)/pkg/project.buildTimestamp=${BUILDTIMESTAMP}' \
-  -X '$(shell go list .)/pkg/project.gitSHA=${GITSHA1}' \
-  -X '$(shell go list .)/pkg/project.version=${VERSION}'
+  -X '$(shell go list .)/pkg/project.gitSHA=${GITSHA1}'
 .DEFAULT_GOAL := build
 
 .PHONY: build build-darwin build-linux
 ## build: builds a local binary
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -55,39 +64,58 @@ $(PACKAGE_DIR)/$(APPLICATION)-v$(VERSION)-%-amd64.tar.gz: $(APPLICATION)-v$(VERS
...
+	@echo "====> $@"
 	go test -ldflags "$(LDFLAGS)" -race ./...
 
 .PHONY: build-docker
 ## build-docker: builds docker image to registry
 build-docker: build-linux
+	@echo "====> $@"
 	docker build -t ${APPLICATION}:${VERSION} .
 
+.PHONY: lint-chart
```
ubergesundheit, line 107: In which context is this bein run? After clone running `make lint-chart` does fail like this on my machine:

```
$ make lint-chart
====> lint-chart
rm -rf /tmp/kubectl-gs-test
mkdir -p /tmp/kubectl-gs-test/helm
cp -a ./helm/kubectl-gs /tmp/kubectl-gs-test/helm/
cp: cannot stat './helm/kubectl-gs': No such file or directory
make: *** [Makefile:114: lint-chart] Error 1
```

axbarsan: Thank you for pointing that out

It's not ran in any way in this case. That's used for apps and operators, I believe, in order to verify that the helm chart is valid.

I opened an issue to only make the script present for the flavours that need it
https://github.com/orgs/giantswarm/projects/163#card-42409142

kopiczko: Does it hurt to have it here?

ubergesundheit: no

axbarsan: It doesn't hurt, but that's why we have different flavours, so we'd be able to only include what we need for each of them

PR #119, opened 2020-07-24
[1/7] get clusters: add output utilities

axbarsan:

These utilities will be used by all `get *` commands.


pkg/output/nameprinter.go
```diff
@@ -0,0 +1,54 @@
...
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/cli-runtime/pkg/printers"
+)
+
+// PrintResourceNames prints the name of a resource, or if it's
+// a list, it will print the names of the nested resources.
+//
+// It can print deeply nested list objects names.
+func PrintResourceNames(out io.Writer, resource runtime.Object) error {
```
axbarsan, line 16: This is necessary because `cli-runtime` utils don't support printing names for `List` types.

pkg/output/output.go
```diff
@@ -0,0 +1,22 @@
+package output
+
+const (
+	TypeDefault        = ""
```
axbarsan, line 4: These are the output types supported by `cli-runtime` print flags. Keeping them here so we wouldn't have to pass strings around

axbarsan:
> Somehow I feel like we're writing some code that should already exist somewhere though, given that kubectl has to do the same type of stuff I imagine?

Please keep in mind that `kubectl` was written before `cli-runtime` was written, and not all `kubectl` functionality is actually using the `cli-runtime` utilities. So some features don't match.

`cli-runtime` does indeed have an utility for printing resource names. We're using it [here](https://github.com/giantswarm/kubectl-gs/blob/fdefb30ee8dddd338cba7af5d4f7354d9a6311d1/pkg/output/nameprinter.go#L22). However, if you'll try to use it with List types, you'll get this error:

image

In order to make printing List types work, I wrote an utility that [extracts](https://github.com/giantswarm/kubectl-gs/blob/fdefb30ee8dddd338cba7af5d4f7354d9a6311d1/pkg/output/nameprinter.go#L33) resources out of List types, and [runs the name printer](https://github.com/giantswarm/kubectl-gs/blob/fdefb30ee8dddd338cba7af5d4f7354d9a6311d1/pkg/output/nameprinter.go#L24) on each one of them.

PR #121, opened 2020-07-24
[3/7] get clusters: add k8s client abstraction

axbarsan:

This is the brain of our plugin. This is the client that interacts handles all the requests towards the K8s API.

It can be used by specific services to interact with the K8s API. It uses the same interface as the client found in our operators. As for exact service implementation, you'll see that in PR 5/7


pkg/data/client/client_fake.go
```diff
@@ -0,0 +1,11 @@
+package client
+
+func NewFakeClient(config Config) (*Client, error) {
```
axbarsan, line 3: This fake client will be used in tests

PR #123, opened 2020-07-27
[4/7] get clusters: add utilities for dealing with config flags

axbarsan:

This PR adds utilities for:
* Guessing the provider based on the current context's k8s api URL (it takes into account flag overriding, such as `--context`, as well)
* Creating a new K8s API client based on the config flags.

The config flags come from `cli-runtime`.

Why is this necessary? Because this code is needed by 99,9% of the commands that interact with the K8s api, so we want to keep the number of code lines that we have to duplicate as little as possible.


pkg/commonconfig/commonconfig.go
```diff
@@ -0,0 +1,65 @@
...
+
+func (cc *CommonConfig) GetProvider() (string, error) {
+	config, err := cc.configFlags.ToRESTConfig()
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	var provider string
+	switch {
+	case strings.Contains(config.Host, key.ProviderAWS):
```
oponder, line 35: Might we run into problems when/if we have an installation called `outlaws` or `jigsaws` that happen to be non AWS installations 😂 

https://www.thefreedictionary.com/words-containing-aws

axbarsan: Good call! 👍 Improved the logic and added some tests for this

PR #124, opened 2020-07-27
[5/7] get clusters: add cluster service

axbarsan:

This PR adds:
* The service for listing clusters (by ID or not, by namespace, by provider), including a mock service for testing.
* A utility for creating a fake kubeconfig for testing

#### Why?
We want to be able to list clusters, but also want to be close to the `kubectl` experience. So we also want to be able to provide a cluster ID and display that cluster in the table or specify a specific namespace to list clusters from, all by being aware of the provider.

We only support modern clusters (`v5`): AWS and Azur


pkg/data/client/error.go
```diff
@@ -4,11 +4,11 @@ import (
 	"github.com/giantswarm/microerror"
 )
 
-var InvalidConfigError = &microerror.Error{
-	Kind: "InvalidConfigError",
+var invalidConfigError = &microerror.Error{
```
axbarsan, line 7: Discovered a mistake in my previous PR. This is supposed to be private, as mentioned in [fmt](https://github.com/giantswarm/fmt/blob/master/go/errors.md#custom-errors)

pkg/data/domain/cluster/aws.go
```diff
@@ -0,0 +1,94 @@
...
+	var err error
+
+	options := &runtimeClient.ListOptions{
+		Namespace: namespace,
+	}
+
+	var clusterIDs map[string]bool
+	{
+		// The CAPI clusters have the labels. They can be used for filtering.
+		apiClusters := &capiv1alpha2.ClusterList{}
```
axbarsan, line 23: Getting Cluster API clusters first, since this have labels, and can be used for filtering

pkg/data/domain/cluster/aws.go
```diff
@@ -0,0 +1,94 @@
...
+			if _, exists := clusterIDs[cluster.Name]; exists {
+				cluster.TypeMeta = infrastructurev1alpha2.NewAWSClusterTypeMeta()
+				clusters = append(clusters, cluster)
+			}
+		}
+		clusterList.Items = clusters
+	}
+
+	{
+		clusterList.APIVersion = "v1"
```
axbarsan, line 57: These need to be added manually because of the way `cli-runtime` printers work. Without them, we get errors over errors over errors over errors 😬

pkg/data/domain/cluster/azure.go
```diff
@@ -0,0 +1,110 @@
...
+		if err != nil {
+			return nil, microerror.Mask(err)
+		} else if len(clusterList.Items) == 0 {
+			return nil, microerror.Mask(noResourcesError)
+		}
+	}
+
+	var objKey runtimeClient.ObjectKey
+	for _, cluster := range clusterList.Items {
+		// FIXME(axbarsan): Remove once the name is stored in
```
axbarsan, line 35: The whole configmap/cluster annotation story will be gone once https://github.com/giantswarm/giantswarm/issues/12363 is done. After that, the configmap will not be necessary anymore, the name will be stored in an annotation by default.

pkg/data/domain/cluster/common.go
```diff
@@ -0,0 +1,90 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/giantswarm/kubectl-gs/internal/key"
+)
+
+const (
+	defaultNamespace = "default"
+)
+
+func (s *Service) Get(ctx context.Context, options *GetOptions) (runtime.Object, error) {
```
axbarsan, line 16: This brings all the service methods together, so we can use a single method in the command runner

pkg/data/domain/cluster/common.go
```diff
@@ -0,0 +1,90 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+
+	"github.com/giantswarm/kubectl-gs/internal/key"
+)
+
+const (
+	defaultNamespace = "default"
+)
+
+func (s *Service) Get(ctx context.Context, options GetOptions) (runtime.Object, error) {
```
axbarsan, line 16: This brings all the service methods together, so we can use a single method in the command runner

pkg/data/domain/cluster/error.go
```diff
@@ -0,0 +1,41 @@
...
+var invalidConfigError = &microerror.Error{
+	Kind: "invalidConfigError",
+}
+
+// IsInvalidConfig asserts invalidConfigError.
+func IsInvalidConfig(err error) bool {
+	return microerror.Cause(err) == invalidConfigError
+}
+
+var noResourcesError = &microerror.Error{
```
axbarsan, line 16: Listing clusters (without an ID) is handled with a custom error because in certain situations we have to handle it differently:
* when using default output type (no `--output`), the process should return the code 0 (no error)
* when using any other output type (e. g. `--output json`), the process should return a non 0 code (error)

pkg/data/domain/cluster/service.go
```diff
@@ -0,0 +1,29 @@
+package cluster
+
+import (
+	"github.com/giantswarm/microerror"
+
+	"github.com/giantswarm/kubectl-gs/pkg/data/client"
+)
+
+var _ Interface = &Service{}
```
axbarsan, line 9: This construct helps us check during compilation if the `Service` struct satisfies the `Interface` interfac

pkg/data/domain/cluster/service_fake.go
```diff
@@ -0,0 +1,53 @@
...
+)
+
+var _ Interface = &FakeService{}
+
+type FakeService struct {
+	service *Service
+	storage []runtime.Object
+}
+
+func NewFakeService(storage []runtime.Object) *FakeService {
```
axbarsan, line 20: We'll use this fake service for testing. It uses the official k8s testing client, so resources are actually verified and created locally.

test/kubeconfig/kubeconfig.go
```diff
@@ -0,0 +1,36 @@
+package kubeconfig
+
+import (
+	"k8s.io/client-go/tools/clientcmd"
+	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+)
+
+func CreateFakeKubeConfig() clientcmd.ClientConfig {
```
axbarsan, line 8: This is useful for mocking `cli-runtime` config flags in tests

PR #158, opened 2020-09-28
Use organization namespace for CAPZ Azure cluster

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/12171

Analog of https://github.com/giantswarm/api/pull/1120

Now the CAPZ `AzureCluster` is created in the `org-<organization-name>` namespace, and the other CRs are still in the `default` namespace


whites11 (approved):
sounds reasonable

fiunchinho (approved):

PR #162, opened 2020-09-30
Add support for using custom releases branch

axbarsan:
Currently, you can only use releases that are in the `master` branch of `giantswarm/releases`. This PR adds a new flag which lets you specify whichever branch you want to use. This is very useful for testing.

### DIFF
--- a/cmd/template/cluster/flag.go
+++ b/cmd/template/cluster/flag.go
@@ -30,15 +30,16 @@ const (
 	flagAzurePublicSSHKey = "azure-public-ssh-key"
 
 	// Common.
-	flagClusterID = "cluster-id"
-	flagDomain    = "domain"
-	flagMasterAZ  = "master-az"
-	flagName      = "name"
-	flagOutput    = "output"
-	flagOwner     = "owner"
-	flagRegion    = "region"
-	flagRelease   = "release"
-	flagLabel     = "label"
+	flagClusterID     = "cluster-id"
+	flagDomain        = "domain"
+	flagMasterAZ      = "master-az"
+	flagName          = "name"
+	flagOutput        = "output"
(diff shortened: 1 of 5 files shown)


cmd/template/cluster/flag.go
```diff
@@ -85,6 +87,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.Region, flagRegion, "", "Installation region (e.g. eu-central-1 or westeurope).")
 	cmd.Flags().StringVar(&f.Release, flagRelease, "", "Tenant cluster release.")
 	cmd.Flags().StringSliceVar(&f.Label, flagLabel, nil, "Tenant cluster label.")
+	cmd.Flags().StringVar(&f.ReleaseBranch, flagReleaseBranch, "", "Release branch to use.")
```
fiunchinho, line 90: should the default be `master`?

axbarsan: We have defaulting here https://github.com/giantswarm/kubectl-gs/blob/e2c4967512482be13b529f64f3ac0d4520cb1d47/pkg/release/release.go#L43

Do you think we need to specify the default here, as well?

fiunchinho: I don't know. I was thinking from the customer point of view. Without the default here, the customer won't know the branch, or will think it needs to pass a branch (since the default is empty). WDYT?

axbarsan: I also specified the branch in the flag defaults. We shouldn't rely on the external `pkg/release` for defaulting.

fiunchinho: Sounds good! Thanks.

fiunchinho (approved):

PR #163, opened 2020-09-30
Store all Azure resources in organization namespace

axbarsan:
Looks like I understood it incorrectly the first time around. I thought that just the `AzureCluster` resource needs to be stored in the organization-specific namespace, but it turns out that all resources must be stored in that namespace. This PR addresses that.

### DIFF
--- a/cmd/template/cluster/runner.go
+++ b/cmd/template/cluster/runner.go
@@ -24,7 +24,6 @@ import (
 
 const (
 	clusterCRFileName        = "clusterCR"
-	organizationNamespaceFmt = "org-%s"
 )
 
 type runner struct {
--- a/cmd/template/cluster/runner.go
+++ b/cmd/template/cluster/runner.go
@@ -2,6 +2,7 @@ package cluster
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
 	"sort"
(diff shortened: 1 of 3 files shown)


cmd/template/cluster/runner.go
```diff
@@ -22,7 +23,8 @@ import (
 )
 
 const (
-	clusterCRFileName = "clusterCR"
+	clusterCRFileName        = "clusterCR"
+	organizationNamespaceFmt = "org-%s"
```
kopiczko, line 27:
```suggestion
```

cmd/template/cluster/runner.go
```diff
@@ -96,6 +98,10 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		if err != nil {
 			return microerror.Mask(err)
 		}
+
+		if r.flag.Provider == key.ProviderAzure {
+			config.Namespace = fmt.Sprintf(organizationNamespaceFmt, config.Owner)
```
kopiczko, line 103:
```suggestion
			config.Namespace = fmt.Sprintf("org-%s", config.Owner)
```

kopiczko (approved):

cmd/template/cluster/runner.go
```diff
@@ -22,7 +23,7 @@ import (
 )
 
 const (
-	clusterCRFileName = "clusterCR"
+	clusterCRFileName        = "clusterCR"
```
fiunchinho, line 26: I think some linting is required here?

fiunchinho (approved):

PR #164, opened 2020-09-30
Use correct 'exp.cluster.x-k8s.io/v1alpha3' api version for machine pools

axbarsan:
Damn it

### DIFF
--- a/cmd/template/cluster/provider/azure.go
+++ b/cmd/template/cluster/provider/azure.go
@@ -20,7 +20,7 @@ import (
 
 const (
 	serviceNetworkCIDR  = "172.31.0.0/16"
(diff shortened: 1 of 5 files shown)


cmd/template/cluster/provider/azure.go
```diff
@@ -20,7 +20,7 @@ import (
 
 const (
 	serviceNetworkCIDR  = "172.31.0.0/16"
-	defaultMasterVMSize = "Standard_D4_v3"
+	defaultMasterVMSize = "Standard_D4s_v3"
```
axbarsan, line 23: Also change the default VM size to the same one in Happa

fiunchinho (approved):

PR #170, opened 2020-10-01
Display full error output when fetching the installation info fails

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/12842

### Preview

image


ubergesundheit (approved):
Code looks good. Unsure if the related issue is worded poorly and if this is really resolving that.

pkg/installation/info.go
```diff
@@ -20,10 +20,10 @@ type installationInfo struct {
 func getInstallationInfo(httpClient *http.Client, apiUrl string) (installationInfo, error) {
 	res, err := httpClient.Get(apiUrl) // #nosec G107
 	if err != nil {
-		return installationInfo{}, microerror.Maskf(cannotGetInstallationInfo, "make sure you're connected to the internet")
+		return installationInfo{}, microerror.Maskf(cannotGetInstallationInfo, "Make sure you're connected to the internet and that the Giant Swarm API is up and running\n%v", err.Error())
```
kopiczko, line 23: Capitalizing the message would break the convetions we have so far.

Maybe we can avoid that by capitalizing first letter during `microerror.Pretty`?

.gitignore
```diff
@@ -1,2 +1,3 @@
 kubectl-gs*
 bin-dist
+.DS_Store
```
kopiczko, line 3: You can add this to your global gitignore https://gist.github.com/subfuzion/db7f57fff2fb6998a16c

kopiczko (approved):

PR #176, opened 2020-10-02
Store Azure node pool resources in the organization namespace

axbarsan:
Same as for clusters, but I thought for some reason that node pools are different


whites11 (approved):
LGTM

tuommaki (approved):

PR #179, opened 2020-10-06
azure v5: remove node pool autoscaler support

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/13649

Remove autoscaler support from node pools, since we want to postpone the feature for later.
Also, set provider-specific defaults (`3` min, `3` max on Azure, or `3` min, `10` max on AWS).

### Preview

<details>
<summary>Trying to use unequal node counts</summary>

image

</details>

<details>
<summary>Node count flags</summary>

image

</details>


cmd/template/nodepool/flag.go
```diff
@@ -121,14 +130,26 @@ func (f *flag) Validate() error {
...
-	}
-	if f.NodesMin < 1 {
-		return microerror.Maskf(invalidFlagError, "--%s must be > 0", flagNodesMin)
-	}
-	if f.NodesMin > f.NodesMax {
-		return microerror.Maskf(invalidFlagError, "--%s must be <= --%s", flagNodesMin, flagNodesMax)
+
+	{
+		// Validate scaling.
+		if f.NodesMax < 1 {
```
tuommaki, line 136: Parameter description mentions default values for these, but how does defaulting work compared to this validation?

axbarsan: I moved the defaulting here, so I could have different defaults for each provider https://github.com/giantswarm/kubectl-gs/blob/6df4611abf93eeb5be9d54f43c87c82c1df7a5cc/cmd/template/nodepool/runner.go#L39

So basically, if the `--provider` flag is `aws`, then the defaults are `min 3` and `max 10`, but if the flag is `azure`, then the defaults are `min 3` and `max 3`

tuommaki: Ah, right. `Validate()` gets called _after_ that defaulting. I missed that earlier. :+1:

tuommaki (approved):

PR #188, opened 2020-10-12
Allow using inactive release versions for templating clusters

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/13581

This is really useful if you want to test version upgrades.


ubergesundheit:
Hope this won't enable users to create clusters with v4 releases.

axbarsan:
> Hope this won't enable users to create clusters with v4 releases.

It doesn't, that validation is done here

https://github.com/giantswarm/kubectl-gs/blob/3d0ac87be6584c49472d7eb8e32d107ae1d99777/pkg/release/release.go#L149

PR #189, opened 2020-10-12
Normalize organization name to match company requirements

axbarsan:
Analog of https://github.com/giantswarm/api/pull/1146

We actually have some requirements for organization names inside our company. (such as using `_` to separate words, or only having lowercase characters), and this PR makes sure that the namespaces we use for creating clusters and nodepools are formatted properly.

### DIFF
--- a/internal/key/key.go
+++ b/internal/key/key.go
@@ -1,6 +1,7 @@
 package key
 
 import (
+	"fmt"
 	"math/rand"
 	"regexp"
 	"strconv"
@@ -18,6 +19,8 @@ const (
 	IDChars = "023456789abcdefghijkmnopqrstuvwxyz"
 	// IDLength represents the number of characters used to create a cluster ID.
 	IDLength = 5
+
+	organizationNamespaceFormat = "org-%s"
 )
 
 func GenerateID() string {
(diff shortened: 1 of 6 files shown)


kopiczko (approved):

internal/key/key.go
```diff
@@ -83,3 +88,9 @@ func ReadSecretYamlFromFile(fs afero.Fs, path string) (map[string][]byte, error)
 
 	return rawMap, nil
 }
+
+func OrganizationNamespaceFromName(name string) string {
+	name = fmt.Sprintf(organizationNamespaceFormat, normalize.AsDNSLabelName(name))
```
tuommaki, line 93: I'd propose to do this other way round:
```suggestion
	name = normalize.AsDNSLabelName(fmt.Sprintf(organizationNamespaceFormat, name))
```

That way the max length trimming of normalization works correctly :slightly_smiling_face:

tuommaki (approved):
Proposal for minor change, in general looks good.

internal/key/key.go
axbarsan: Very good call 👍🏻

PR #195, opened 2020-10-15
Removed ssh key parameter for azure

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/13783

This PR remove the `public ssh key` parameter that was mandatory on azure.
This is not used any more.


axbarsan (approved):
Thank you 👍🏻

PR #196, opened 2020-10-15
Add "kubectl-gs validate apps" command

oponder:
Towards: https://github.com/giantswarm/roadmap/issues/152

This adds the "kubectl-gs validate apps" command as specced here: 

https://github.com/giantswarm/giantswarm/blob/master/content/docs/architecture-specs-adrs/specs/kubectl-gs/validate-apps.md

I need a bit of help with:

- [x] Stopping the debug output from app-operator packages
- [x] Customizing the 'supported output formats'. It should really just be "table" or "report". 
        `Output format. One of: json|yaml|name|go-template|go-template-file|template|templatefile|jsonpath|jsonpath-file.`


oponder:
@stone-z @giantswarm/sig-security : I'm running into lint rule G107 https://securego.io/docs/rules/g107.html

In this case, I want to download a file based on a URL that is stored in a CRD. This URL is technically user input, especially when we consider the `helm-stable` catalog. So it is definitely  not guaranteed 'safe' input.

That said.. i don't see any way around using that URL, it's pretty essential to this feature.

We expect the downloaded file to be json, and the main thing we do with is unmarshal it. Unless there are exploits in json.Unmarshal, i think this is ok to disable this lint rule.

Any thoughts?

oponder:
FYI @giantswarm/sig-security I've opted to skip the lint rule for that line. Based on my understanding, it is ok to do so in this case.

stone-z:
Hey @oponder sorry I missed your ping here. You're right it would be hard to avoid in this case, I think the exception is fine. Just please leave a note about why it's necessary in the code

oponder:
Some important findings after testing a schema I generated for nginx-ingress-controller-app using `helm schema gen` (https://github.com/karuppiah7890/helm-schema-gen) against a live installation.

It looks like the json schema validator is running into an issue validating an integer. For some reason it think the type is unknown.

To investigate: 
- [ ]  replacing the json schema validation library.
- [ ] finding the bug in the actual validation library.
- [ ] or maybe theres a bug in the kubectl-gs code  somehow.

oponder:
Not sure how to start writing tests here... Will try to take inspiration from other tests in this repo. But @giantswarm/chapter-front-end , perhaps you can already provide some feedback on the code so far.

pkg/data/domain/app/validate.go
```diff
@@ -0,0 +1,418 @@
...
+	// If the namespace is empty, set it to "default".
+	if namespace == "" {
+		namespace = defaultNamespace
+	}
+
+	// BUT if we want all namespaces, set it to 'metav1.NamespaceAll', aka ""
+	// again so the client gets all namespaces.
+	if options.AllNamespaces {
+		namespace = metav1.NamespaceAll
+	}
```
axbarsan, line 42: In the `get clusters` command, we now do this defaulting at the command level

https://github.com/giantswarm/kubectl-gs/blob/78d6a3cad6e3f531b00622773e62c6d7a02db119/cmd/get/clusters/runner.go#L77

pkg/data/domain/app/validate.go
```diff
@@ -0,0 +1,418 @@
...
+		if err != nil {
+			return "", nil, microerror.Mask(err)
+		}
+	}
+
+	// Use the helm binary to fetch and untar the chart tarball.
+	// Why? There are actually a lot of little security details that go into
+	// unpacking this tarball. Checkout https://github.com/helm/helm/blob/master/pkg/chart/loader/archive.go#L101
+	url := findTarballURL(index.Entries[app.Spec.Name], app.Spec.Version)
+	tmpDir, err := ioutil.TempDir("", app.Spec.Name+"-"+app.Spec.Version+"-")
```
axbarsan, line 172:
```suggestion
	tmpDir, err := ioutil.TempDir("", fmt.Sprintf("%s-%s-", app.Spec.Name, app.Spec.Version))
```

pkg/data/domain/app/validate.go
```diff
@@ -0,0 +1,418 @@
...
+import (
+	"context"
+	"io/ioutil"
+	"net/http"
+	"os/exec"
+	"path"
+
+	applicationv1alpha1 "github.com/giantswarm/apiextensions/v2/pkg/apis/application/v1alpha1"
+	"github.com/xeipuuv/gojsonschema"
+	"gopkg.in/yaml.v3"
```
axbarsan, line 12: FMT states that we should use `sigs.k8s.io/yaml` instead.

https://github.com/giantswarm/fmt/blob/master/go/yaml_marshalling.md

pkg/data/domain/app/service.go
```diff
@@ -0,0 +1,55 @@
...
+	Client *client.Client
+	Logger micrologger.Logger
+}
+
+type Service struct {
+	client        *client.Client
+	valuesService *values.Values
+
+	catalogFetchResults map[string]CatalogFetchResult
+	schemaFetchResults  map[string]SchemaFetchResult
```
axbarsan, line 24: The data service should only be used to fetch resources from the k8s API (hence the package name `data`). It shouldn't include additional business logic, such as caching results (here), or the app validation logic inside of `pkg/data/domain/app/validate.go`.

This business logic should live inside the command that uses it, or even better, inside a separate package in the `pkg` folder.

pkg/data/domain/app/service.go
```diff
@@ -0,0 +1,55 @@
...
+}
+
+func New(config Config) (Interface, error) {
+	var err error
+
+	if config.Client == nil {
```
axbarsan, line 30: The logger should also be checked here

axbarsan (changes requested):
The PR looks good in general, but I have a few concerns

pkg/data/domain/app/service.go
oponder: Oh makes sense!

I moved this code now to `pkg/app` . There are a few places where it calls out to the k8s API, but it's pretty straightforward. Not sure it makes sense to extract that then into a specific data package just yet.

oponder:
@axbarsan thx for the thorough review. I think I addressed your points so far. Could you take a look again?

pkg/app/validate.go
```diff
@@ -0,0 +1,407 @@
+package app
```
axbarsan, line 1: Apologies for not being clear enough. This is what I meant with the refactor:

* Fetching apps and app catalogs should still belong to the `data` package (e.g. `pkg/data/domain/app` for app-specific k8s functionality)
* Validating apps would be done here, in the `app` package, with the list of apps provided as input for the validation functions.

However, that might be a lot of work to do at once at this point, and maybe having to rework the logic, so we can maybe do it incrementally in other PRs.
But as for this PR, it believe it would be worth moving the k8s fetching logic to the `data` package, and providing the `app` and `appcatalog` data services as input for these validation functions. That would make it easier to rework later.

Please let me know what you think

oponder: OK let me give it a try. I thought to skip doing that for now since the actual fetching logic is not as complex as in the cluster creation case (theres no provider specific things to be aware of in this case) 

but since it makes sense from a code organisation point of view, I'll give it a go

go.mod
```diff
@@ -3,38 +3,51 @@ module github.com/giantswarm/kubectl-gs
...
+	gopkg.in/yaml.v3 v3.0.0-20200313102051-9f266ea9e77c
 	k8s.io/api v0.18.9
 	k8s.io/apiextensions-apiserver v0.18.9
 	k8s.io/apimachinery v0.18.9
 	k8s.io/cli-runtime v0.18.9
 	k8s.io/client-go v0.18.9
+	k8s.io/utils v0.0.0-20200912215256-4140de9c8800 // indirect
 	sigs.k8s.io/cluster-api v0.3.10
-	sigs.k8s.io/cluster-api-provider-azure v0.4.9
+	sigs.k8s.io/cluster-api-provider-azure v0.4.8
```
axbarsan, line 38:
```suggestion
	sigs.k8s.io/cluster-api-provider-azure v0.4.9
```

axbarsan (commented):
LGTM after fixing the Azure CAPI dependency version

oponder:
Alright, this should be good to go now. ping @axbarsan can I get an official approval?

axbarsan (approved):
Thank you for cleaning everything up 👍🏻

PR #202, opened 2020-10-23
Removed the `--domain` flag on `Azure`

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/13916

This PR removes the `--domain` flag for the `template` command on `azure only`.
The CR fields that needed that value are left blank and populated using a mutating webhook.

The rationale behind this is that there is no configurability of the `domain` field, so there is no need to add a flag that has no more that one specific option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 ### Changed
 
 - Removed availability zones for `GermanyWestCentral` in `Azure`.
+- Removed the `--domain` flag on `Azure`.
 
 ## [0.9.0] - 2020-10-16
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
-### Changed
+### Removed
 
 - Removed availability zones for `GermanyWestCentral` in `Azure`.
 - Removed the `--domain` flag on `Azure`.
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 ### Changed
 
 - Removed availability zones for `GermanyWestCentral` in `Azure`.
+- Removed the `--domain` flag on `Azure`.
```
axbarsan, line 13: These should be under the `### Removed` section

cmd/template/cluster/flag.go
```diff
@@ -105,8 +105,14 @@ func (f *flag) Validate() error {
 
 		return nil
 	}
-	if f.Domain == "" {
-		return microerror.Maskf(invalidFlagError, "--%s must not be empty", flagDomain)
+	{
+		// Validate domain.
+		switch f.Provider {
```
axbarsan, line 110: Should we throw an error if an user tries to use it on `Azure`?

CHANGELOG.md
whites11: Right, thanks

cmd/template/cluster/flag.go
fiunchinho: Yeah, maybe it makes sense to tell the customer that it's not allowed on Azure.

whites11: Added

fiunchinho (approved):
LGTM

axbarsan (approved):
LGTM! 👍🏻  Please don't forget about the docs
https://github.com/giantswarm/docs/blob/master/src/content/reference/kubectl-gs/template-cluster.md

PR #205, opened 2020-10-28
Add support for listing clusters in all namespaces, or using a namespace from the current context

axbarsan:
2 things:

* Add support for using a specific namespace defined for a kubectl context (you can define that using `kubectl config set-context --current --namespace=<insert-namespace-name-here>`)
* Add support for the `--all-namespaces` flag, which allows listing clusters in all namespaces (aka all organizations)

### DIFF
--- a/cmd/get/clusters/runner.go
+++ b/cmd/get/clusters/runner.go
@@ -68,7 +68,6 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 	{
 		options := cluster.GetOptions{
 			Provider:  r.provider,
-			Namespace: metav1.NamespaceDefault,
 		}
 		{
 			if len(args) > 0 {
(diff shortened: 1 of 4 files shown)


T-Kukawka (approved):
🚀

cmd/get/clusters/runner.go
```diff
@@ -67,14 +68,21 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 	var resource runtime.Object
 	{
 		options := cluster.GetOptions{
-			Provider: r.provider,
+			Provider:  r.provider,
+			Namespace: metav1.NamespaceDefault,
```
fiunchinho, line 72: is this necessary if we are going to overwrite it with either `NamespaceAll` or whatever is in the config file?

fiunchinho (approved):
small comment, LGTM otherwise

cmd/get/clusters/runner.go
axbarsan: That was there in case the type cast failed... But I just realized that we don't actually need that type cast anymore, so we don't need the default either

PR #213, opened 2020-11-10
Removed the `--region` flag from the `kubectl-gs template` commands. …

whites11:
The region flag was used to calculate the base domain of the installation and to do some validation around availability zones on the client side.

In any case the value of such flag is not arbitrary, it has to match the control plane's region.

For the aformentioned reason, this PR removes the --region flag from the `template nodepool` and `template cluster` commands.

Needless to say this requires the provider admission controller to set the needed fields during cluster and nodepool creation.

This is already done in `azure`, will wait for @giantswarm/team-firecracker-engineers  to be ready before merging


axbarsan (approved):
👍🏻

fiunchinho (approved):

anvddriesch (approved):
`region` defaulting in AWS was released so removing this is fine with us.

PR #218, opened 2020-11-13
App data package

oponder:
Hi Adrian! This is towards your comment in the kubectl-gs app validate branch: https://github.com/giantswarm/kubectl-gs/pull/196#discussion_r521180036

If you have a chance, does this look like what you had in mind? If so I'll do the same for the remaining k8s api call for app catalogs in the validate package.


axbarsan (approved):
Really nice! Thank you for taking the time to do this

PR #219, opened 2020-11-17
Create app catalog data pkg and use it in validate calls.

oponder:
One more extraction that I'll merge into the feature branch.

After that I'd like to merge the feature, and do further refactoring and adding tests in a new PR.


axbarsan (approved):
Spotless 👍🏻

PR #221, opened 2020-11-18
Extract interaction with helm binary into own package.

oponder:
Also extract the helm binary interaction into its own package.

Took some time since the other PR wasn't approved yet so did this in the meanwhile.

### DIFF
--- a/pkg/helmbinary/service.go
+++ b/pkg/helmbinary/service.go
@@ -1,0 +1,53 @@
+package helmbinary
+
+import (
+	"context"
+	"io/ioutil"
+	"os/exec"
+
+	"github.com/giantswarm/kubectl-gs/pkg/data/client"
+	"github.com/giantswarm/microerror"
+)
+
+var _ Interface = &Service{}
+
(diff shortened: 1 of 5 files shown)


axbarsan (approved):

pkg/helmbinary/service.go
```diff
@@ -0,0 +1,64 @@
...
+	tmpDir, err = ioutil.TempDir("", "kubectl-gs-validate-apps-")
+	if err != nil {
+		return "", microerror.Mask(err)
+	}
+
+	// nosec is applied here, the URL in question can be traced back to the
+	// app catalog index entry. We verified this is actually a URL earlier in this
+	// method.
+	cmd := exec.Command("helm", "pull", parsedURL.String(), "--untar", "-d", tmpDir) // #nosec G204
+	_, err = cmd.CombinedOutput()
```
kopiczko, line 58: I just skimmed trough it and it looks weird. By default the output is discarded so it would be clearer to do `cmd.Run()` instead.

PR #235, opened 2020-12-03
Remove credential flag

pipo02mix:
Since it defaults to credential secret name based on org, now we should not set `credential` anymore as it could be error prone. @whites11 is azure defaulting credential too? if not we need to wait till it does

towards https://github.com/giantswarm/roadmap/issues/198

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+- Removed the `--credential` flag, now it is managed by admission controller.
+
 ## [0.14.0] - 2020-11-24
 
 ### Added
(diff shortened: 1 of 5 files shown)


whites11:
> Since it defaults to credential secret name based on org, now we should not set `credential` anymore as it could be error prone. @whites11 is azure defaulting credential too? if not we need to wait till it does
> 
> towards [giantswarm/roadmap#198](https://github.com/giantswarm/roadmap/issues/198)

According to intellij That's used by AWS only.
If you want to be 100% sure just try creating a cluster with `kubectl-gs` (your branch) on `gremlin` in org `429-testing` (that is BYOC) and check if it works.

anvddriesch (approved):
Not sure why the build still fails, but looks good to me.

whites11 (approved):

PR #236, opened 2020-12-03
Add support for non-standard GS API URLs

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/14545

Because some installations have non-standard GS API URLs, we need to somehow support logging in for all of them.
This PR includes a solution that'a bit hacky, trying out multiple common API URLs and seeing if it works.

I don't consider it a big issue for now, since we want to have a different service providing installation information in the future, instead of the GS API.


oponder (approved):
LGTM!

PR #254, opened 2020-12-21
[GET NP 3/3]: Implement the 'get nodepools' command

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/14854

This PR includes the command implementation and tests for the `get nodepools` command. Don't be scared of the large number of changes, they're mostly goldenfile test snapshots.

PR for documentation is linked in the attached issue.

### Preview

<details><summary>Get nodepools</summary>

image

</details>

<details><summary>Get nodepool by ID</summary>

image

</details>

<details><summary>Get nodepools by cluster ID</summary>

image

</details>

<details><summary>Get nodepools, no nodepools available</summary>

image

</details>

<details><summary>Get nodepools by ID, no nodepool found</summary>

image

</details>

<details><summary>Help</summary>

image

</details>


cmd/get/nodepools/flag.go
```diff
@@ -0,0 +1,36 @@
...
+type flag struct {
+	AllNamespaces bool
+	ClusterID     string
+
+	config genericclioptions.RESTClientGetter
+	print  *genericclioptions.PrintFlags
+}
+
+func (f *flag) Init(cmd *cobra.Command) {
+	cmd.Flags().BoolVarP(&f.AllNamespaces, flagAllNamespaces, "A", false, "If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even if specified with --namespace.")
```
pipo02mix, line 22: is no already defined the kubectl-specific ones?

axbarsan: Unfortunately no... we do the same for the `get clusters` command, as well
https://github.com/giantswarm/kubectl-gs/blob/bc1138ebb28cd0e1125329961f240520c5dd387b/cmd/get/clusters/flag.go#L20

pipo02mix (approved):
LGTM nice work!

oponder (approved):
LGTM!

marians:
How is the output table sorted? I can't find any client side sorting.

Guessing from the example output, it's by node pool ID, right? Looks to me like if I had several node pools in the same cluster, they would be spread over the table. Not sure I would like that as a user.

axbarsan:
> How is the output table sorted? I can't find any client side sorting.
> 
> Guessing from the example output, it's by node pool ID, right? Looks to me like if I had several node pools in the same cluster, they would be spread over the table. Not sure I would like that as a user.

There is no client-side sorting implemented, but sorting by cluster ID makes a lot of sense. I added it now 👍🏻

marians:
Can we influence the "Usage" part of the help output? To me, `<nodepool-id>` indicates that the node pool ID is a mandatory argument. It should be wrapped in `[]` to indicate it's optional.

```nohighlight
...

Usage:
  kubectl gs get nodepools <nodepool-id> [flags]

...
```

cmd/get/nodepools/command.go
```diff
@@ -0,0 +1,98 @@
...
+- NODES MIN/MAX: Node pool autoscaler settings (if supported).
+- NODES DESIRED: The total number of nodes that the node pool should have.
+- NODES READY: The number of nodes in the node pool that are actually ready.
+- DESCRIPTION: User friendly description for the node pool.`
+
+	examples = `  # List all node pools you have access to
+  kgs get nodepools
+
+  # Get one specific nodepool by its ID
+  kgs get nodepools 3f01a
```
marians, line 39: Feels better in singular form when used with ID argument:

```suggestion
  kgs get nodepool 3f01a
```

marians:
Wha is the `CONDITION` column always `n/a`? Testing on gorilla (AWS) so far.

axbarsan:
> Why is the `CONDITION` column always `n/a`? Testing on gorilla (AWS) so far.

Node pool conditions are unsupported on AWS at the moment. Here's how they look on the latest Azure release:

image

marians (approved):
Cool! Added nitpicks, but nothing blocking.

PR #259, opened 2021-01-12
[GET NP 1/3]: Add support for computing provider capabilities

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/14854

This PR adds support for checking if a specific feature is supported on a cluster, based on its provider and release version. The reason for doing it is so we could have an easy way of hiding and displaying information, based on a feature being supported or not.

For the moment, I only added `Autoscaling` and `Node Pool Conditions` to the list.

### DIFF
--- a/internal/feature/feature.go
+++ b/internal/feature/feature.go
@@ -1,0 +1,44 @@
+package feature
+
+import (
+	"strings"
+
+	"github.com/blang/semver"
+)
+
+// Service is used to compute different provider capabilities.
+type Service struct {
+	provider string
+	features map[string]Feature
+}
+
+func New(provider string) *Service {
+	s := &Service{
+		provider: provider,
+		features: map[string]Feature{
+			Autoscaling:        autoscaling,
+			NodePoolConditions: nodePoolConditions,
+		},
+	}
+
+	return s
+}
+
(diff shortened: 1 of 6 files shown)


internal/feature/feature.go
```diff
@@ -0,0 +1,44 @@
...
+			Autoscaling:        autoscaling,
+			NodePoolConditions: nodePoolConditions,
+		},
+	}
+
+	return s
+}
+
+// Supports checks if a certain feature is supported or not on a given release version.
+func (s *Service) Supports(featureName string, releaseVersion string) bool {
```
oponder, line 28: Cool data structure and implementation

oponder (approved):
LGTM :+1:

internal/feature/feature.go
```diff
@@ -0,0 +1,44 @@
...
+	if !exists {
+		return false
+	}
+
+	capability, exists := feature[s.provider]
+	if !exists {
+		return false
+	}
+
+	releaseVersion = strings.TrimPrefix(releaseVersion, "v")
```
ubergesundheit, line 39: This thing again :D 

not for this PR but we maybe should introduce a function to normalize the version for internal use

ubergesundheit (approved):

PR #261, opened 2021-01-13
Remove the mention of the unexistent 'create cluster' command, and suggest 'template cluster' instead

axbarsan:
If there are no cluster resources to be printed, we suggest creating one using the `create cluster` command, which doesn't exist. Let's suggest `template cluster` instead.


oponder (approved):
LGTM 👍

marians:
Thanks!

PR #265, opened 2021-01-20
Validate context before considering the user logged in

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/15200

This PR makes sure that if we're trying to re-use the currently selected kube context for logging in, we validate that it's configured properly, so that none of the properties are missing.

### Preview

<details>
<summary>Logging in with an installation codename/kubectl context (have already been logged in before)</summary>

image

</details>

<details>
<summary>Logging in without any arguments (just renewing the token)</summary>

image

</details>

### DIFF
--- a/cmd/login/auth.go
+++ b/cmd/login/auth.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"time"
 
 	gooidc "github.com/coreos/go-oidc"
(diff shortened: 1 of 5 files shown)


cmd/login/auth.go
```diff
@@ -265,15 +271,32 @@ func switchContext(ctx context.Context, k8sConfigAccess clientcmd.ConfigAccess,
...
+	}
+	for _, k := range keys {
+		if len(provider.Config[k]) == 0 {
+			return microerror.Mask(invalidAuthConfigurationError)
+		}
 	}
 
-	if !kubeconfig.IsKubeContext(config.CurrentContext) {
-		return config.CurrentContext, false
+	_, err := url.ParseRequestURI(provider.Config[Issuer])
```
axbarsan, line 296: Also make sure that the auth token issuer is a valid URL.

ubergesundheit (approved):
Seems reasonable

PR #266, opened 2021-01-20
Use absolute paths for k8s certs

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/15265

Pretty straightforward

### DIFF
--- a/cmd/login/auth.go
+++ b/cmd/login/auth.go
@@ -202,7 +202,7 @@ func storeCredentials(k8sConfigAccess clientcmd.ConfigAccess, i *installation.In
 		config.CurrentContext = contextName
 	}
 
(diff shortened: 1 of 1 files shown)


cmd/login/auth.go
```diff
@@ -202,7 +202,7 @@ func storeCredentials(k8sConfigAccess clientcmd.ConfigAccess, i *installation.In
 		config.CurrentContext = contextName
 	}
 
-	err = clientcmd.ModifyConfig(k8sConfigAccess, *config, true)
+	err = clientcmd.ModifyConfig(k8sConfigAccess, *config, false)
```
axbarsan, line 205: `true` - relative paths
`false` - absolute paths

ubergesundheit (approved):
nice and simple :D

PR #260, opened 2021-01-12
[GET NP 2/3]: Add data domain for node pools

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/14854

This PR adds the K8s API logic for the `get nodepools` command.

What's different compared to the clusters logic is that we'll use a wrapper around node pool resources. This allows us to easily get all the required information from multiple resources (such as the CAPI `MachinePool`, and the CAPZ `AzureMachinePool`), but still be able to easily print a singe resource as YAML or JSON.

Hopefully the comments below do a good job in explaining this.

PR #268, opened 2021-01-27
Return the CAPI Cluster resource from the 'get clusters' command

axbarsan:
Fixes https://github.com/giantswarm/giantswarm/issues/13748

This PR makes use of the [same type of abstraction](https://github.com/giantswarm/kubectl-gs/pull/260) that we used with the `get nodepools` command, so we can make use of whatever resources we want when printing information (such as both CAPI `Cluster` and custom `AWSCluster` resource on AWS).

### DIFF
--- a/cmd/get/clusters/testdata/print_list_of_azure_clusters_json_output.golden
+++ b/cmd/get/clusters/testdata/print_list_of_azure_clusters_json_output.golden
@@ -5,12 +5,13 @@
     "items": [
         {
             "kind": "Cluster",
-            "apiVersion": "v1alpha3",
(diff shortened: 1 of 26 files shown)


cmd/get/clusters/testdata/print_list_of_azure_clusters_json_output.golden
```diff
@@ -5,12 +5,13 @@
     "items": [
         {
             "kind": "Cluster",
-            "apiVersion": "v1alpha3",
+            "apiVersion": "cluster.x-k8s.io/v1alpha3",
```
axbarsan, line 8: This was wrong before

oponder (approved):
LGTM

PR #272, opened 2021-02-09
Use spark CR as bootstrap config ref when templating a new machinepool

axbarsan:
Same as https://github.com/giantswarm/api/pull/1223


fiunchinho (changes requested):
Changelog needs to be updated.

fiunchinho (approved):

PR #273, opened 2021-02-09
Add missing AzureMachine/AzureCluster fields for CAPZ validation

axbarsan:
Towards https://github.com/giantswarm/giantswarm/issues/15570

I have no idea what I'm doing


cmd/template/cluster/provider/azure.go
```diff
@@ -85,6 +85,18 @@ func newAzureClusterCR(config ClusterCRsConfig) *capzv1alpha3.AzureCluster {
 		},
 		Spec: capzv1alpha3.AzureClusterSpec{
 			ResourceGroup: config.ClusterID,
+			NetworkSpec: capzv1alpha3.NetworkSpec{
+				APIServerLB: capzv1alpha3.LoadBalancerSpec{
+					Name: fmt.Sprintf("%s-%s", config.ClusterID, "public-lb"),
```
tuommaki, line 90: Mostly fine, but these names should match with what we deploy in Azure. They are currently defined main `deployment` template here: https://github.com/giantswarm/azure-operator/blob/master/service/controller/azureconfig/handler/deployment/template/main.json#L1100-L1117

Therefore e.g. this LB name should be `<clusterid>-PublicLoadBalancer`.

cmd/template/cluster/provider/azure.go
```diff
@@ -125,8 +137,9 @@ func newAzureMasterMachineCR(config ClusterCRsConfig) *capzv1alpha3.AzureMachine
 				},
 			},
 			OSDisk: capzv1alpha3.OSDisk{
-				OSType:     "Linux",
-				DiskSizeGB: int32(50),
+				OSType:      "Linux",
+				CachingType: "None",
```
fiunchinho, line 141: I think we use `ReadWrite` as caching type.

tuommaki (approved):
Seems good to me :+1:

fiunchinho (approved):

PR #285, opened 2021-03-05
Extends cluster ID alphanumeric validation

JosephSalisbury:
Without this PR:
```
$ kgs template cluster --provider=aws --cluster-id=fooba
Error: Invalid flag: --cluster-id must be alphanumeric
$ kgs template cluster --provider=aws --cluster-id=12345
Error: Invalid flag: --cluster-id must be alphanumeric
```
This confused me somewhat, as `fooba` and `12345` are _technically_ alphanumeric (just without digits or letters, respectively).

With this PR:
```
$ kgs template cluster --provider=aws --cluster-id=fooba
Error: Invalid flag: --cluster-id must contain at least one number
$ kgs template cluster --provider=aws --cluster-id=12345
Error: Invalid flag: --cluster-id must contain at least one digit
```
Which I think is a bit clearer about what we want.

### DIFF
--- a/cmd/template/cluster/flag.go
+++ b/cmd/template/cluster/flag.go
@@ -76,13 +76,19 @@ func (f *flag) Validate() error {
 			return microerror.Maskf(invalidFlagError, "--%s must be length of %d", flagClusterID, key.IDLength)
 		}
 
-		matched, err := regexp.MatchString("^([a-z]+|[0-9]+)$", f.ClusterID)
-		if err == nil && matched {
-			// strings is letters only, which we also avoid
-			return microerror.Maskf(invalidFlagError, "--%s must be alphanumeric", flagClusterID)
+		matchedLettersOnly, err := regexp.MatchString("^[a-z]+$", f.ClusterID)
+		if err == nil && matchedLettersOnly {
+			// strings is letters only, which we avoid
+			return microerror.Maskf(invalidFlagError, "--%s must contain at least one number", flagClusterID)
 		}
 
-		matched, err = regexp.MatchString("^[a-z0-9]+$", f.ClusterID)
+		matchedNumbersOnly, err := regexp.MatchString("^[0-9]+$", f.ClusterID)
+		if err == nil && matchedNumbersOnly {
+			// strings is numbers only, which we avoid
+			return microerror.Maskf(invalidFlagError, "--%s must contain at least one digit", flagClusterID)
+		}
+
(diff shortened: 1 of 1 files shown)


JosephSalisbury:
I am assuming here we want to avoid cluster IDs with only digits, give a shout if that's not the case.

QuentinBisson:
Some cluster ids are only digits though

cmd/template/cluster/flag.go
```diff
@@ -76,13 +76,19 @@ func (f *flag) Validate() error {
...
 		}
 
-		matched, err := regexp.MatchString("^([a-z]+|[0-9]+)$", f.ClusterID)
-		if err == nil && matched {
-			// strings is letters only, which we also avoid
-			return microerror.Maskf(invalidFlagError, "--%s must be alphanumeric", flagClusterID)
+		matchedLettersOnly, err := regexp.MatchString("^[a-z]+$", f.ClusterID)
+		if err == nil && matchedLettersOnly {
+			// strings is letters only, which we avoid
+			return microerror.Maskf(invalidFlagError, "--%s must contain at least one number", flagClusterID)
```
QuentinBisson, line 82:
```suggestion
			return microerror.Maskf(invalidFlagError, "--%s must contain at least one digit", flagClusterID)
```

JosephSalisbury:
Yeah, with ya. I'll update so it's better messaging if it's only letters.

JosephSalisbury:
actually, master disallows digit only cluster ids. @marians can you confirm if we do or do not want to allow digit only cluster ids?

marians (approved):

marians:
Currently, digit-only cluster IDs may break some expectations, but not sure.

TheoBrigitte (approved):
LGTM

kopiczko:
AFAICT CAPI doesn't allow clusters that start with dight, should we prevent that too?

marians:
We normally want that sort of validation in the admission controllers

kopiczko:
Then why do we have any validation here?

marians:
That is a very good question

JosephSalisbury:
okay i'm just gonna merge here for now, i'm just improving error messages, not changing behaviour :D

JosephSalisbury:
@marians do we have tracking somewhere for general improvements on cluster ID management when it comes to CAPI?

JosephSalisbury:
(e.g: that we could possibly get rid of it with CAPI, or reduce our requirements)

PR #287, opened 2021-03-09
Checks all matching pods for CAPI controllers

JosephSalisbury:
We also match webhooks with our labels, so check all relevant pods for the CAPI controllers.


JosephSalisbury:
@axbarsan i was apparently v. naughty and never added this to CHANGELOG, so i'm sneaking it in now :D

axbarsan (approved):

PR #288, opened 2021-03-09
Print program errors to stderr

axbarsan:
This makes sure that if you're piping the output of the program into a stream, you'll still see error messages in your terminal.

Before
```
kubectl gs some-command-that-fails > /dev/null
```

After
```
kubectl gs some-command-that-fails > /dev/null
Error: Invalid flag: --provider must be either aws or azure
exit status 1
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


kopiczko (approved):

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
```
kopiczko, line 10: Isn’t that fixed?

PR #289, opened 2021-03-10
Use PKCE auth flow (without a secret)

axbarsan:
Needs https://github.com/giantswarm/dex-app/pull/54


oponder (approved):
LGTM

marians:
Let's release this as a new major version, as it causes a bump in the UX.

axbarsan:
> Let's release this as a new major version, as it causes a bump in the UX.

Not sure if we need to bump to `v2.0` just for an internal authentication change. It does not change the way you run the login command.

PR #297, opened 2021-03-19
Use embed instead of pkger for HTML templates

marians:
This makes access to the templates files easier and removes the pkger dependency.

I checked but couldn't find any developer docs in the repo where pkger is mentioned.


marians:
@axbarsan Seems like the project is built in an older Go version. 

Do we have to update https://github.com/giantswarm/kubectl-gs/blob/master/.circleci/config.yml#L4 ?

I can't even tell whether that's architect or architect-orb, so I could need some help with chosing the right version number to use. Also wondering whether it should be with or without the `v` prefix behind the `@`.

kopiczko (approved):

axbarsan:
> @axbarsan Seems like the project is built in an older Go version. "embed" requires Go 1.16.
> 
> Do we have to update [https://github.com/giantswarm/kubectl-gs/blob/master/.circleci/config.yml#L4](https://github.com/giantswarm/kubectl-gs/blob/master/.circleci/config.yml?rgh-link-date=2021-03-19T07%3A51%3A16Z#L4) ?
> 
> I can't even tell whether that's architect or architect-orb, so I could need some help with chosing the right version number to use. Also wondering whether it should be with or without the `v` prefix behind the `@`.

We need `architect` newer than `v2.0.0`. Also, we need the changes to the GitHub workflows [in this PR](https://github.com/giantswarm/devctl/pull/227), so we could create working executable binaries

kopiczko:
I guess you'll have to also update https://github.com/giantswarm/kubectl-gs/blob/817e8ae90911d6c1d0c2de7bfc3f473b11eb66c8/go.mod#L3

ubergesundheit (approved):

PR #300, opened 2021-03-30
Implement `get orgs` command

marians:
This PR adds a `get orgs` command.

Compared to `kubectl get organizations` it offers two benefits:

- The command also works for non admins (users which only have  `get` permissions to specific Organization resources).
- The output shows both the organization name and the organization namespace.

## Output preview

```nohighlight
TODO: add help output
```

## Help

```
NAME                  ORG NAMESPACE             AGE
anti-party            org-anti-party            20d
antonia               org-antonia               311d
cedric                org-cedric                257d
dmitry                org-dmitry                151d
george                org-george                47d
giantswarm            org-giantswarm            489d
gollum-subscription   org-gollum-subscription   552d
gremlin               org-gremlin               487d
herve                 org-herve                 47d
ljakimczuk            org-ljakimczuk            322d
marcel                org-marcel                277d
marian                org-marian                384d
nikola                org-nikola                353d
party-org             org-party-org             379d
ross                  org-ross                  347d
sandy                 org-sandy                 361d
team-rainbow          org-team-rainbow          185d
theo                  org-theo                  266d
tobias                org-tobias                70d
vaclav-v              org-vaclav-v              12d
```

### DIFF
--- a/pkg/data/domain/organization/common.go
+++ b/pkg/data/domain/organization/common.go
@@ -2,13 +2,14 @@ package organization
 
 import (
 	"context"
+	"sort"
+
 	securityv1alpha1 "github.com/giantswarm/apiextensions/v6/pkg/apis/security/v1alpha1"
 	"github.com/giantswarm/microerror"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sort"
 )
 
 func (s *Service) Get(ctx context.Context, options GetOptions) (Resource, error) {
@@ -49,9 +50,9 @@ func (s *Service) getByName(ctx context.Context, name string) (Resource, error) 
 
 		o.Organization = omitManagedFields(org)
 		o.Organization.GetObjectKind().SetGroupVersionKind(schema.GroupVersionKind{
-			Group: securityv1alpha1.SchemeGroupVersion.Group,
+			Group:   securityv1alpha1.SchemeGroupVersion.Group,
 			Version: securityv1alpha1.SchemeGroupVersion.Version,
-			Kind: "Organization",
+			Kind:    "Organization",
 		})
 	}
 
@@ -132,8 +133,6 @@ func (s *Service) getAllPermittedOrgResources(ctx context.Context) (Resource, er
 func (s *Service) getByNames(ctx context.Context, names []string) (*Collection, error) {
 
 	orgCollection := &Collection{}
-
-
 
 	for _, name := range names {
 		resource, err := s.getByName(ctx, name)
--- a/pkg/data/domain/organization/common.go
+++ b/pkg/data/domain/organization/common.go
@@ -1,0 +1,169 @@
+package organization
+
+import (
+	"context"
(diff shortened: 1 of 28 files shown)


pkg/data/domain/organization/spec.go
```diff
@@ -0,0 +1,57 @@
...
+	Object() runtime.Object
+}
+
+// Organization gives access to the actual Organization resource.
+type Organization struct {
+	*securityv1alpha1.Organization
+}
+
+func (o *Organization) Object() runtime.Object {
+	return o
```
axbarsan, line 31: This should do it 👍🏻 

```suggestion
	return o.Organization
```

marians:
UX feedback from inside Giant Swarm:

> it could be nice to see how many clusters exist in each organization, and the corresponding namespace.
> I don't think the age matters.

tfussell:
Tested against gamma:
```
kubectl-gs get orgs
NAME                  AGE
conformance-testing   3d18h
multi-project         4d15h
single-gs2            4d15h
single-gs3            4d15h
```

pkg/data/domain/organization/common.go
```diff
@@ -0,0 +1,168 @@
...
+	sort.Strings(orgNames)
+
+	return s.getByNames(ctx, orgNames)
+}
+
+func (s *Service) getByNames(ctx context.Context, names []string) (*Collection, error) {
+
+	orgCollection := &Collection{}
+
+	for _, name := range names {
```
vvondruska, line 137: Loading Organization CRs for a list of provided org names.
Sequential load appears to be fast enough even with larger sizes of the org name list (10s in our context)
Some resources in kubectl-gs are pulled sequentially, some in parallel, so parallel approach can also be considered here.

pkg/data/domain/organization/common.go
```diff
@@ -0,0 +1,168 @@
...
+			Version: securityv1alpha1.SchemeGroupVersion.Version,
+			Kind:    "Organization",
+		})
+	}
+
+	return o, nil
+}
+
+func (s *Service) getAll(ctx context.Context) (Resource, error) {
+	accessReview, err := s.getSelfSubjectAccessReview(ctx)
```
vvondruska, line 63: First we check the SelfSubjectAccessReview to see if the user has permissions to list Organization CRs.

If the user does have sufficient permissions, then we load the list of Organization CRs

Otherwise we use SelfSubjectResourceRules to get a list of resources available to the user, we filter out names of resources of the "Organization" type, and we pull CR for each individual org name, which are visible to the user

vvondruska:
Finished required features of get orgs:
- Enabled non-admin users to list organizations
- Provided additional details in the table output of the `get orgs` command
- Added support for additional output types
- Added tests
- Fixed compatibility of the org service with the WC certificate generating service

marians:
FYI I'll test in a bit

marians:
I noticed that we are still using the alias `kgs` here in the usage. We should change that to `kubectl gs`, as we decided to not use `kgs` any more. (I must have caused this as I opened this branch so long ago.)

Apart from that, looks good!

anvddriesch (commented):
LGTM! I'm going to try it locally, too. :)
Please add a changelog entry.

anvddriesch:
Works fine for me. Tests look good, too.
I just have a nitpick. IMO `org` should be the alias and `organization` should be the name we see in the usage of `get`.
```
  clusters    Display one or many clusters
  nodepools   Display one or many node pools
  orgs        Display one or many organizations
  releases    Display one or more releases
```

anvddriesch (approved):

PR #307, opened 2021-04-07
No longer rely on the REST API for fetching installation information

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/16657

This PR refactors the existing installation info service to use `athena`, instead of the GS REST API. It also adds a small Graph QL client with the minimal functionality required to talk to `athena`.


pkg/graphql/client.go
```diff
@@ -0,0 +1,89 @@
...
+	HttpClient *http.Client
+	Url        string
+}
+
+type ClientImpl struct {
+	httpClient *http.Client
+	url        string
+}
+
+var _ Client = (*ClientImpl)(nil)
```
axbarsan, line 21: Check that `ClientImpl` implements the `Client` interface.

This is better than
```go
var _ Client = &ClientImpl{}
```
because it does no memory allocations.

pkg/graphql/response.go
```diff
@@ -0,0 +1,8 @@
+package graphql
+
+import "encoding/json"
+
+type responseBody struct {
+	Data   *json.RawMessage        `json:"data"`
```
axbarsan, line 6: With the `json.RawMessage` type, we can leave the `data` object unmarshaled, so we can unmarshal it later in the context that calls the client.

For example:
https://github.com/giantswarm/kubectl-gs/blob/e78a4abf48c2284760121938d6a79dbc60a850de/pkg/installation/info.go#L49

pkg/graphql/client.go
corest: I never even saw such code construction in Golang

axbarsan: https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/kubelet/cm/topologymanager/topology_manager.go#L116

This is for checking during compilation if a type implements an interface. If that type doesn't implement the interface correctly, the compilation would fail

marians: If this is not common knowledge, you might want to add a comment, to enlighten the future maintainer :)

marians (approved):
LGTM and works for me

pkg/graphql/client.go
```diff
@@ -0,0 +1,93 @@
+package graphql
```
kopiczko, line 1: Funny: https://github.com/giantswarm/config-controller/blob/master/pkg/github/internal/graphql/client.go

pkg/graphql/response.go
kopiczko: Maybe for inspiration: in my implementation I'm unmarshalling directly to the object provided https://github.com/giantswarm/config-controller/blob/948f76c29f9da500b4f7ab307203cdafbced4458/pkg/github/internal/graphql/client.go#L36

kopiczko (approved):

pkg/graphql/response.go
axbarsan: That's actually quite interesting, and closer to how built-in functions behave. I'll refactor this to match that 👍🏻 

Also, I had no idea you already built a GraphQL client... I would've just copied that 😅

pkg/graphql/client.go
axbarsan: There's a better worded explanation in Effective Go
https://golang.org/doc/effective_go#blank_implements

PR #311, opened 2021-04-13
fix ci build when using architect-orb 2.7.0

njuettner:
PR addresses fix for failing CI https://github.com/giantswarm/kubectl-gs/pull/310

### DIFF
--- a/internal/key/key.go
+++ b/internal/key/key.go
@@ -28,6 +28,7 @@ const (
 func GenerateID() string {
 	compiledRegexp, _ := regexp.Compile("^[a-z]+$")
 
+	/* #nosec G404 */
 	for {
 		letterRunes := []rune(IDChars)
 		b := make([]rune, IDLength)
--- a/pkg/data/client/k8sclient_fake.go
+++ b/pkg/data/client/k8sclient_fake.go
@@ -25,7 +25,7 @@ import (
 	capiv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
 	capiexpv1alpha3 "sigs.k8s.io/cluster-api/exp/api/v1alpha3"
 	"sigs.k8s.io/controller-runtime/pkg/client"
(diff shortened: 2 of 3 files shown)


calvix (approved):

internal/key/key.go
```diff
@@ -28,6 +28,7 @@ const (
 func GenerateID() string {
 	compiledRegexp, _ := regexp.Compile("^[a-z]+$")
 
+	/* #nosec G404 */
```
njuettner, line 31: gosec complaining about using crypto/rand

pkg/data/client/k8sclient_fake.go
```diff
@@ -25,7 +25,7 @@ import (
 	capiv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
 	capiexpv1alpha3 "sigs.k8s.io/cluster-api/exp/api/v1alpha3"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake" //nolint:staticcheck // v0.6.4 has a deprecation on pkg/client/fake that was removed in later versions
```
njuettner, line 28: fake client is being deprecated, we fixed it by ignoring it for now in most of our repos

test/goldenfile/goldenfile.go
```diff
@@ -33,7 +33,7 @@ func (gf *GoldenFile) Read() ([]byte, error) {
 }
 
 func (gf *GoldenFile) Update(data []byte) error {
-	err := ioutil.WriteFile(gf.path, data, 0644)
+	err := ioutil.WriteFile(gf.path, data, 0600)
```
njuettner, line 36: gosec complaining about file permissions, goldenfile is used for tests. should be fine

PR #318, opened 2021-04-22
Fix templating YAML for configmaps and secrets referenced in App and AppCatalog CRs

rossf7:
Fixes a problem I found while working on https://github.com/giantswarm/docs/pull/912

When using `kubectl gs template app` or appcatalog nested values can't be passed because we need to convert to a `map[string]interface{}{}` first.

### DIFF
--- a/internal/key/key.go
+++ b/internal/key/key.go
@@ -81,8 +81,7 @@ func ReadSecretYamlFromFile(fs afero.Fs, path string) ([]byte, error) {
 		return nil, microerror.Mask(err)
 	}
 
-	rawMap := map[string]interface{}{}
-	err = yaml.Unmarshal(data, &rawMap)
+	err = yaml.Unmarshal(data, &map[string]interface{}{})
 	if err != nil {
 		return nil, microerror.Maskf(unmashalToMapFailedError, err.Error())
 	}
--- a/internal/key/key.go
+++ b/internal/key/key.go
@@ -65,7 +65,7 @@ func ReadConfigMapYamlFromFile(fs afero.Fs, path string) (string, error) {
 		return "", microerror.Mask(err)
 	}
 
-	rawMap := map[string]string{}
+	rawMap := map[string]interface{}{}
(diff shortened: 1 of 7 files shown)


internal/key/key.go
```diff
@@ -65,7 +65,7 @@ func ReadConfigMapYamlFromFile(fs afero.Fs, path string) (string, error) {
 		return "", microerror.Mask(err)
 	}
 
-	rawMap := map[string]string{}
+	rawMap := map[string]interface{}{}
```
rossf7, line 68: This is the fix for the nested values.

pkg/template/app/app.go
```diff
@@ -17,7 +17,7 @@ type Config struct {
 }
 
 type SecretConfig struct {
-	Data      map[string][]byte
+	Data      []byte
```
rossf7, line 20: Aligning this with how we template configmaps. The only difference is secrets use a `[]byte` type instead of a string.

internal/key/key.go
```diff
@@ -75,19 +75,19 @@ func ReadConfigMapYamlFromFile(fs afero.Fs, path string) (string, error) {
...
-func ReadSecretYamlFromFile(fs afero.Fs, path string) (map[string][]byte, error) {
+func ReadSecretYamlFromFile(fs afero.Fs, path string) ([]byte, error) {
 	data, err := afero.ReadFile(fs, path)
 	if err != nil {
 		return nil, microerror.Mask(err)
 	}
 
-	rawMap := map[string][]byte{}
+	rawMap := map[string]interface{}{}
 	err = yaml.Unmarshal(data, &rawMap)
```
tomahawk28, line 85: Is it just validating purpose? then what do you think to use an anonymous variable?
```suggestion
	err = yaml.Unmarshal(data, &map[string]interface{}{})
```

axbarsan (approved):
Da real MVP

internal/key/key.go
rossf7: Good catch and yes we only need this for validating. Thx!

PR #323, opened 2021-04-28
Fix the internal prefix composition

pipo02mix:


pkg/installation/installation.go
```diff
@@ -54,7 +54,7 @@ func New(ctx context.Context, fromUrl string) (*Installation, error) {
 		return nil, microerror.Mask(err)
 	}
 
-	k8sInternalAPI := fmt.Sprintf("https://%s.%s", internalAPIPrefix, basePath)
+	k8sInternalAPI := fmt.Sprintf("https://%s-%s", internalAPIPrefix, basePath)
```
pipo02mix, line 57: when I change to base path I did not realize the URL already container `g8s`

axbarsan (approved):

PR #327, opened 2021-04-30
Refactor app data service to return runtime objects

rossf7:

This refactors the app data service to return runtime objects and aligns it with the cluster data service. 
This is prep for adding the `kubectl gs get apps` command.

```sh
$ go run main.go validate apps -n ni0t5

NAMESPACE   NAME                           VERSION   RESULT
ni0t5       cert-exporter                  1.6.1     n/a: app has no values schema
ni0t5       cert-manager                   2.4.3     0 errors
ni0t5       chart-operator                 2.13.0    n/a: app has no values schema
ni0t5       cluster-autoscaler             1.19.2    n/a: app has no values schema
ni0t5       coredns                        1.4.1     n/a: app has no values schema
ni0t5       external-dns                   2.3.0     0 errors
ni0t5       kiam                           1.7.1     n/a: app has no values schema
ni0t5       kube-state-metrics             1.3.1     n/a: app has no values schema
ni0t5       metrics-server                 1.2.2     n/a: app has no values schema
ni0t5       net-exporter                   1.9.3     n/a: app has no values schema
ni0t5       nginx-ingress-controller-app   1.16.1    1 error
ni0t5       node-exporter                  1.7.2     n/a: app has no values schema

$ go run main.go validate app -n ni0t5 nginx-ingress-controller-app -o report

Validated 1 app across 1 namespace

ni0t5 [1 app, 1 error]
  nginx-ingress-controller-app:
    controller.admissionWebhooks.enabled - Invalid type. Expected: boolean, given: string - wrong

$ go run main.go validate apps -n ni0t5 missing-app

Error: Apps.application.giantswarm.io "missing-app" not found
exit status 1
```

### DIFF
--- a/pkg/app/get.go
+++ b/pkg/app/get.go
@@ -1,1 +1,0 @@
-package app
--- a/pkg/app/validate.go
+++ b/pkg/app/validate.go
@@ -2,16 +2,17 @@ package app
 
 import (
 	"context"
+	"fmt"
 	"io/ioutil"
 	"net/http"
 	"path"
 
+	applicationv1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/microerror"
 	"github.com/xeipuuv/gojsonschema"
 	"sigs.k8s.io/yaml"
 
-	applicationv1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/application/v1alpha1"
-	"github.com/giantswarm/microerror"
-
+	"github.com/giantswarm/kubectl-gs/pkg/data/domain/app"
 	appdata "github.com/giantswarm/kubectl-gs/pkg/data/domain/app"
 	appcatalogdata "github.com/giantswarm/kubectl-gs/pkg/data/domain/appcatalog"
 	"github.com/giantswarm/kubectl-gs/pkg/helmbinary"
@@ -53,15 +54,24 @@ func (s *Service) validateByName(ctx context.Context, name, namespace string, cu
 		Name:      name,
 	}
 
-	app, err := s.appDataService.Get(ctx, options)
-	if err != nil {
+	appResource, err := s.appDataService.Get(ctx, options)
+	if app.IsNotFound(err) {
+		return nil, microerror.Maskf(notFoundError, fmt.Sprintf("An app '%s/%s' cannot be found.\n", options.Namespace, options.Name))
+	} else if err != nil {
 		return results, microerror.Mask(err)
 	}
 
-	valuesSchema, schemaValidationResult, err := s.validateApp(ctx, *app, customValuesSchema)
+	var appCR *applicationv1alpha1.App
+
+	switch a := appResource.(type) {
+	case *app.App:
+		appCR = a.CR
+	}
+
+	valuesSchema, schemaValidationResult, err := s.validateApp(ctx, appCR, customValuesSchema)
 	if err != nil {
 		results = append(results, &ValidationResult{
-			App:          *app,
+			App:          appCR,
 			ValuesSchema: "",
 			Err:          microerror.Mask(err),
 		})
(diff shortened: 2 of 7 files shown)


pkg/app/get.go
```diff
@@ -1 +0,0 @@
-package app
```
rossf7, line 1: Just dropping this file since it's empty.

pkg/app/validate.go
```diff
@@ -82,22 +92,33 @@ func (s *Service) validateByName(ctx context.Context, name, namespace string, cu
...
 	var err error
 
-	options := appdata.ListOptions{
+	options := appdata.GetOptions{
 		Namespace:     namespace,
 		LabelSelector: labelSelector,
 	}
 
-	apps, err := s.appDataService.List(ctx, options)
+	appResource, err := s.appDataService.Get(ctx, options)
```
rossf7, line 100: I replaced List with Get to align with the cluster data service.

pkg/data/domain/app/service.go
```diff
@@ -37,40 +37,80 @@ func New(config Config) (Interface, error) {
...
+func (s *Service) Get(ctx context.Context, options GetOptions) (Resource, error) {
+	var resource Resource
+	var err error
+
+	if len(options.Name) > 0 {
+		resource, err = s.getByName(ctx, options.Name, options.Namespace)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	} else {
```
tomahawk28, line 51: I don't think we need `else` statement here, right?

tomahawk28 (commented):
Looks good, only few suggestions from myside.

pkg/data/domain/app/service.go
```diff
@@ -37,40 +37,80 @@ func New(config Config) (Interface, error) {
...
-		LabelSelector: parsedSelector,
-	}
+	return appCollection, nil
+}
 
-	appList := &applicationv1alpha1.AppList{}
-	err = s.client.K8sClient.CtrlClient().List(ctx, appList, o)
-	if err != nil {
-		return nil, microerror.Mask(err)
+func (s *Service) getByName(ctx context.Context, name, namespace string) (Resource, error) {
```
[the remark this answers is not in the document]
rossf7: Order is fixed.

pkg/app/validate.go
```diff
@@ -82,22 +92,33 @@ func (s *Service) validateByName(ctx context.Context, name, namespace string, cu
...
-	} else if len(apps.Items) == 0 {
+	}
+
+	var apps []*applicationv1alpha1.App
+
+	switch a := appResource.(type) {
+	case *app.Collection:
+		for _, appItem := range a.Items {
+			apps = append(apps, appItem.CR)
+		}
```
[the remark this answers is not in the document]
rossf7: Yes much better. I added default statements in both places and now returns an error.

pkg/data/domain/app/service.go
rossf7: Good catch we don't need it.

tomahawk28 (approved):
LGTM

pkg/data/domain/app/service.go
```diff
@@ -37,40 +37,82 @@ func New(config Config) (Interface, error) {
...
+		appCR := &applicationv1alpha1.App{}
+		err = s.client.K8sClient.CtrlClient().Get(ctx, runtimeclient.ObjectKey{
+			Namespace: namespace,
+			Name:      name,
+		}, appCR)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		} else if apierrors.IsNotFound(err) {
+			return nil, microerror.Mask(notFoundError)
+		}
```
axbarsan, line 108: This needs to be the other way around, otherwise `apierrors.IsNotFound()` will never get executed

```suggestion
		if apierrors.IsNotFound(err) {
			return nil, microerror.Mask(notFoundError)
		} else if err != nil
			return nil, microerror.Mask(err)
		}
```

axbarsan (changes requested):
Thank you ❤️
LGTM apart from this small thing

pkg/data/domain/app/service.go
rossf7: Thanks! Nice catch @axbarsan

That's updated PTAL

axbarsan (approved):

PR #329, opened 2021-05-04
Add get apps command

rossf7:

This adds the get apps command.

```sh
$ kubectl gs get apps -h

Display one or many apps

Output columns:

- NAME: Name of the app.
- VERSION: Version of the app.
- LAST DEPLOYED: When the app was last deployed.
- STATUS: Status of the app.

Usage:
  kubectl gs get apps <app-name> [flags]

Aliases:
  apps, app

Examples:
  # List all apps for the current namespace
  kubectl gs get apps

  # Get one app by its name
  kubectl gs get app coredns
...

$ kubectl gs get apps -n dmie3

NAME                 VERSION   LAST DEPLOYED   STATUS
cert-exporter        1.2.3     10d             deployed
cert-manager         2.3.2     10d             deployed
chart-operator       2.5.1     10d             deployed
cluster-autoscaler   1.17.3    10d             deployed
coredns              1.2.0     10d             deployed
external-dns         1.5.0     10d             deployed
kiam                 1.5.0     10d             deployed
kube-state-metrics   1.1.1     10d             deployed
metrics-server       1.1.1     10d             deployed
net-exporter         1.9.0     10d             deployed
node-exporter        1.3.0     10d             deployed
```

### DIFF
--- a/cmd/get/apps/printer.go
+++ b/cmd/get/apps/printer.go
@@ -69,6 +69,8 @@ func getTable(appResource app.Resource) *metav1.Table {
 	}
 
 	switch c := appResource.(type) {
+	case *app.App:
+		table.Rows = append(table.Rows, getAppRow(*c))
 	case *app.Collection:
 		for _, appItem := range c.Items {
 			table.Rows = append(table.Rows, getAppRow(appItem))
--- a/cmd/get/apps/printer.go
+++ b/cmd/get/apps/printer.go
@@ -51,6 +51,11 @@ func (r *runner) printOutput(appResource app.Resource) error {
 	return nil
 }
 
+func (r *runner) printNoMatchOutput() {
+	fmt.Fprintf(r.stdout, "No App CRD found.\n")
+	fmt.Fprintf(r.stdout, "Please check you are accessing a management cluster\n\n")
+}
+
 func (r *runner) printNoResourcesOutput() {
 	fmt.Fprintf(r.stdout, "No apps found.\n")
 	fmt.Fprintf(r.stdout, "To create an app, please check\n\n")
--- a/cmd/get/apps/printer.go
+++ b/cmd/get/apps/printer.go
@@ -1,0 +1,97 @@
+package apps
+
+import (
+	"fmt"
+
+	"github.com/giantswarm/microerror"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/cli-runtime/pkg/printers"
+
+	"github.com/giantswarm/kubectl-gs/pkg/data/domain/app"
+	"github.com/giantswarm/kubectl-gs/pkg/output"
+)
+
+func (r *runner) printOutput(appResource app.Resource) error {
+	var (
+		err      error
+		printer  printers.ResourcePrinter
+		resource runtime.Object
+	)
+
+	switch {
+	case output.IsOutputDefault(r.flag.print.OutputFormat):
+		resource = getTable(appResource)
+		printOptions := printers.PrintOptions{
+			WithNamespace: r.flag.AllNamespaces,
+		}
+		printer = printers.NewTablePrinter(printOptions)
+	case output.IsOutputName(r.flag.print.OutputFormat):
+		resource = appResource.Object()
+		err = output.PrintResourceNames(r.stdout, resource)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+
+		return nil
+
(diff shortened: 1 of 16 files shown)


pkg/output/duration.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/duration"
+)
+
+// TranslateTimestampSince returns the elapsed time since timestamp in
+// human-readable approximation.
```
rossf7, line 11: This is based on translateTimestampSince. So the duration matches the one kubectl shows.

https://github.com/kubernetes/cli-runtime/blob/v0.21.0/pkg/printers/tableprinter.go#L564-L572

pkg/data/domain/app/service.go
```diff
@@ -116,3 +116,10 @@ func (s *Service) getByName(ctx context.Context, namespace, name string) (Resour
 
 	return app, nil
 }
+
+// omitManagedFields removes managed fields to make YAML output easier to read.
+// With Kubernetes 1.21 we can use OmitManagedFieldsPrinter and remove this.
+func omitManagedFields(app *applicationv1alpha1.App) *applicationv1alpha1.App {
+	app.ManagedFields = nil
+	return app
+}
```
rossf7, line 125: OmitManagedFieldsPrinter got added in K8s 1.21. So we can remove this then.

https://github.com/kubernetes/cli-runtime/blob/v0.21.0/pkg/printers/managedfields.go#L26-L30

rossf7:
Tested on gauss.

cmd/get/apps/printer.go
```diff
@@ -0,0 +1,97 @@
...
+func getAppRow(a app.App) metav1.TableRow {
+	if a.CR == nil {
+		return metav1.TableRow{}
+	}
+
+	return metav1.TableRow{
+		Cells: []interface{}{
+			a.CR.Name,
+			a.CR.Status.Version,
+			output.TranslateTimestampSince(a.CR.Status.Release.LastDeployed),
```
axbarsan, line 90: Nice ❤️

pkg/data/domain/app/service.go
axbarsan: 👍🏻

ubergesundheit:
What happens if this command is executed with a WC kubeconfig? It just fails or is there a notice printed, that this probably does not work on a WC?

axbarsan (approved):
Works great for me. Thank you for going the extra mile for the human-readable duration thingy 🥳

cmd/get/apps/runner.go
```diff
@@ -0,0 +1,133 @@
...
+				options.Name = strings.ToLower(args[0])
+			}
+
+			if r.flag.AllNamespaces {
+				options.Namespace = metav1.NamespaceAll
+			} else {
+				options.Namespace, _, err = r.flag.config.ToRawKubeConfigLoader().Namespace()
+				if err != nil {
+					return microerror.Mask(err)
+				}
```
axbarsan, line 90: This logic is also on line #65

rossf7:
> What happens if this command is executed with a WC kubeconfig? It just fails or is there a notice printed, that this probably does not work on a WC?

@ubergesundheit Currently just the CRD missing error is printed. Not sure why there is the throttling message from my test cluster.

Good idea to show a custom message. I'll get that done.


```
go run main.go get apps -n giantswarm
I0504 11:56:34.906668   94876 request.go:621] Throttling request took 1.056960859s, request: GET:https://api.ni0t5.k8s.gauss.eu-west-1.aws.gigantic.io/apis/coordination.k8s.io/v1?timeout=32s
Error: No matches for kind "App" in version "application.giantswarm.io/v1alpha1"
exit status 1
```

rossf7:
```
go run main.go get apps -n giantswarm
I0504 14:13:14.348650   98718 request.go:621] Throttling request took 1.078231816s, request: GET:https://api.ni0t5.k8s.gauss.eu-west-1.aws.gigantic.io/apis/coordination.k8s.io/v1?timeout=32s
No App CRD found.
Please check you are accessing a management cluster
```

@ubergesundheit I added this message. 

The rate limiter error is still there but it includes the API server URL which is kind of useful. So I think we can live with it. WDYT?

ubergesundheit:
> The rate limiter error is still there but it includes the API server URL which is kind of useful. So I think we can live with it. WDYT?

Fine with me :)

ubergesundheit (approved):

PR #330, opened 2021-05-04
Refactor appcatalog data service to return runtime objects

rossf7:

Same as https://github.com/giantswarm/kubectl-gs/pull/327 but for the appcatalog data service.

As well as getting the appcatalog CR it also supports getting the appcatalogentry CRs for a catalog. This is so `kubectl gs get appcatalog <catalog>` can also show the apps in the catalog.


pkg/data/domain/appcatalog/spec.go
```diff
@@ -4,16 +4,70 @@ import (
...
+
+// Collection wraps a list of appcatalogs.
+type Collection struct {
+	Items []AppCatalog
+}
+
 // GetOptions are the parameters that the Get method takes.
 type GetOptions struct {
-	Name string
+	LabelSelector string
```
axbarsan, line 25: What do you think about using the `labels.Selector` type? This way we require the command (or the caller) to parse and validate the input, in the same way we do with command line flags.

axbarsan (approved):
LGTM

pkg/data/domain/appcatalog/spec.go
rossf7: Thanks @axbarsan I like that so the input can be validated. Done.

PR #333, opened 2021-05-06
Handle x509.UnknownAuthorityError specifically when connecting to athena

marians:
Towards https://github.com/giantswarm/giantswarm/issues/16307

This PR should make the error message a bit mor specific when the athena certificate is not trusted.

I know, this is a very specific piece of error handling. However, I did it to debug a problem and figured I might as well suggest it for inclusion. WDYT?

### DIFF
--- a/pkg/graphql/client.go
+++ b/pkg/graphql/client.go
@@ -3,8 +3,10 @@ package graphql
 import (
 	"bytes"
 	"context"
+	"crypto/x509"
 	"encoding/json"
 	"net/http"
+	"net/url"
 
 	"github.com/giantswarm/microerror"
 )
@@ -66,6 +68,11 @@ func (c *ClientImpl) ExecuteQuery(ctx context.Context, query string, variables m
 
 	res, err := c.httpClient.Do(req)
 	if err != nil {
+		if urlErr, ok := err.(*url.Error); ok {
+			if x509Err, ok := urlErr.Err.(x509.UnknownAuthorityError); ok {
+				return microerror.Maskf(unknownAuthorityError, "certificate issued by %q", x509Err.Cert.Issuer.CommonName)
+			}
+		}
 		return microerror.Mask(err)
 	}
 	defer res.Body.Close()
--- a/pkg/graphql/error.go
+++ b/pkg/graphql/error.go
@@ -16,6 +16,16 @@ func IsInvalidConfig(err error) bool {
 	return microerror.Cause(err) == invalidConfigError
 }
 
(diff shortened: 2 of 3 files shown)


pkg/graphql/client.go
```diff
@@ -66,6 +68,11 @@ func (c *ClientImpl) ExecuteQuery(ctx context.Context, query string, variables m
 
 	res, err := c.httpClient.Do(req)
 	if err != nil {
+		if urlErr, ok := err.(*url.Error); ok {
```
axbarsan, line 71: Can you please use `errors.As()` instead? There are examples here
https://golang.org/src/errors/errors.go

marians: Oh cool

marians: Done

pkg/graphql/client.go
```diff
@@ -68,8 +69,10 @@ func (c *ClientImpl) ExecuteQuery(ctx context.Context, query string, variables m
 
 	res, err := c.httpClient.Do(req)
 	if err != nil {
-		if urlErr, ok := err.(*url.Error); ok {
-			if x509Err, ok := urlErr.Err.(x509.UnknownAuthorityError); ok {
+		var urlErr *url.Error
+		if errors.As(err, &urlErr) {
+			var x509Err x509.UnknownAuthorityError
```
axbarsan, line 74: I think this should be a pointer type.

```suggestion
			var x509Err *x509.UnknownAuthorityError
```

Also, I don't think this functionality should be in the GraphQL client, as its unrelated to it, but rather it should be in the installation info package. 

https://github.com/giantswarm/kubectl-gs/blob/6d9b79edd21404cd23fe3e54d2bcb076af891d3d/pkg/installation/info.go#L45

pkg/graphql/client.go
```diff
@@ -66,6 +69,13 @@ func (c *ClientImpl) ExecuteQuery(ctx context.Context, query string, variables m
 
 	res, err := c.httpClient.Do(req)
 	if err != nil {
+		var urlErr *url.Error
+		if errors.As(err, &urlErr) {
+			var x509Err x509.UnknownAuthorityError
+			if errors.As(urlErr, &x509Err) {
```
axbarsan, line 75: I think you were matching this property initially

```suggestion
			if errors.As(urlErr.Err, &x509Err) {
```

marians:
Doesn't make sense to invest more time here. Thanks anyway for looking into it.

PR #335, opened 2021-05-07
Fix China AZs

paurosello:
China only has 2 AZs so we should open this a bit. In other regions, although we do not recommend it, 2AZ also works

### DIFF
--- a/cmd/template/cluster/flag.go
+++ b/cmd/template/cluster/flag.go
@@ -4,8 +4,6 @@ import (
 	"net"
 	"regexp"
 
-	"github.com/mpvl/unique"
-
 	"github.com/giantswarm/microerror"
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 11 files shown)


calvix (approved):
lgtm if tested

cmd/template/cluster/flag.go
```diff
@@ -112,8 +112,8 @@ func (f *flag) Validate() error {
 		// Validate Master AZs.
 		switch f.Provider {
 		case key.ProviderAWS:
-			if len(f.MasterAZ) != 0 && len(f.MasterAZ) != 1 && len(f.MasterAZ) != 3 {
-				return microerror.Maskf(invalidFlagError, "--%s must be set to either one or three availability zone names", flagMasterAZ)
+			if len(f.MasterAZ) < 1 && len(f.MasterAZ) > 3 {
```
axbarsan, line 115: Not sure if the length can be both less than 1 and more than 3 at the same time 😬 

```suggestion
			if len(f.MasterAZ) < 1 || len(f.MasterAZ) > 3 {
```

axbarsan (changes requested):

axbarsan:
Please don't forget about updating `CHANGELOG.md` 👍🏻

anvddriesch:
LGTM!
Just make sure this works correctly with 3 AZ names showing up in the CR like this:
```
availabilityZones:
- AZ1
- AZ2
- AZ2
```

paurosello:
> LGTM!
> Just make sure this works correctly with 3 AZ names showing up in the CR like this:
> 
> ```
> availabilityZones:
> - AZ1
> - AZ2
> - AZ2
> ```

Oh true! we should not allow 2 AZ, just enable repeating AZ as the master needs to be either 1 or 3 always

axbarsan (approved):
LGTM after fixing CI 🌟

PR #345, opened 2021-05-12
Update k8s dependencies to v0.18.18

marians:


axbarsan (commented):
Is there a reason behind this? Almost all of our projects are using `v0.18.9`

marians:
Well, it's a patch update. Dependabot made me aware that it is available, and I'm updating as possible, like we normally do with dependencies.

axbarsan (approved):
LGTM if tested

PR #346, opened 2021-05-12
login: add support for having a timeout for calling the auth proxy

axbarsan:
This PR adds a limit of `1 minute` for the `login` command to call the authentication proxy, so the app won't remain open accidentally if login was not finalized.

Also, the auth proxy port is no longer hardcoded, it is automatically set to an available one.

### Preview

<details>
<summary>Error message, if the timeout is reached</summary>

image

</details>

### DIFF
--- a/pkg/callbackserver/callbackserver.go
+++ b/pkg/callbackserver/callbackserver.go
@@ -16,6 +16,7 @@ type Config struct {
 	Port        int
 	RedirectURI string
 }
+
 type CallbackServer struct {
 	port        int
 	redirectURI string
--- a/pkg/callbackserver/callbackserver.go
+++ b/pkg/callbackserver/callbackserver.go
@@ -72,7 +72,12 @@ func (cs *CallbackServer) Run(ctx context.Context, callback CallbackFunc) (inter
 			Handler: mux,
 		}
 	}
-	defer server.Shutdown(ctx)
+	defer func() {
+		err := server.Shutdown(ctx)
+		if err != nil {
+			errorCh <- err
+		}
+	}()
 
 	go func() {
 		err := server.ListenAndServe()
--- a/pkg/callbackserver/callbackserver.go
+++ b/pkg/callbackserver/callbackserver.go
@@ -72,12 +72,6 @@ func (cs *CallbackServer) Run(ctx context.Context, callback CallbackFunc) (inter
 			Handler: mux,
 		}
 	}
-	defer func() {
-		err := server.Shutdown(ctx)
-		if err != nil {
-			errorCh <- err
-		}
-	}()
 
 	go func() {
 		err := server.ListenAndServe()
@@ -100,7 +94,16 @@ func (cs *CallbackServer) Run(ctx context.Context, callback CallbackFunc) (inter
 		break
 	}
 
-	return result, microerror.Mask(origErr)
+	err := server.Shutdown(ctx)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	if origErr != nil {
+		return nil, microerror.Mask(origErr)
+	}
+
+	return result, nil
 }
 
 func (cs *CallbackServer) Port() int {
--- a/pkg/callbackserver/callbackserver.go
+++ b/pkg/callbackserver/callbackserver.go
@@ -72,16 +72,16 @@ func (cs *CallbackServer) Run(ctx context.Context, callback CallbackFunc) (inter
 			Handler: mux,
 		}
 	}
+	defer server.Shutdown(ctx)
 
-	// Start the server.
-	go (func(s *http.Server) {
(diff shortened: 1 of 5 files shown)


pkg/callbackserver/callbackserver.go
```diff
@@ -2,58 +2,121 @@ package callbackserver
```
axbarsan, line 2: Cleaned up this implementation a bit

pkg/callbackserver/callbackserver.go
```diff
@@ -2,58 +2,121 @@ package callbackserver
...
+	var result interface{}
+	var origErr error
+	select {
+	case result = <-resultCh:
+		break
+	case origErr = <-errorCh:
+		break
+	case <-ctx.Done():
+		origErr = microerror.Mask(timedOutError)
+		break
```
axbarsan, line 95: The context timeout was exceeded, let's bounce

pkg/callbackserver/callbackserver.go
```diff
@@ -2,58 +2,121 @@ package callbackserver
...
+	}
+
+	ln, err := net.Listen("tcp", addr.String())
+	if err != nil {
+		return -1, microerror.Mask(err)
 	}
+
+	port := ln.Addr().(*net.TCPAddr).Port
+
+	return port, nil
```
axbarsan, line 123: Find a random open port

pkg/callbackserver/callbackserver.go
```diff
@@ -2,58 +2,123 @@ package callbackserver
...
+	}
 
 	// Start the server.
-	go startServer(s)
+	go (func(s *http.Server) {
+		err := s.ListenAndServe()
+		if errors.Is(err, http.ErrServerClosed) {
+			// All good.
+		} else if err != nil {
+			panic(err)
```
kopiczko, line 82: Maybe:

```suggestion
			errorCh <- err
```

pkg/callbackserver/callbackserver.go
```diff
@@ -2,58 +2,123 @@ package callbackserver
...
+	case origErr = <-errorCh:
+		break
+	case <-ctx.Done():
+		origErr = microerror.Mask(timedOutError)
+		break
+	}
 
-	// Shutdown the server.
-	err := s.Shutdown(context.Background())
+	err := server.Shutdown(ctx)
```
kopiczko, line 98: Looks like a candidate for `defer`.

pkg/callbackserver/callbackserver.go
```diff
@@ -2,58 +2,123 @@ package callbackserver
...
+func findAvailablePort() (int, error) {
+	addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
+	if err != nil {
+		return -1, microerror.Mask(err)
+	}
+
+	ln, err := net.Listen("tcp", addr.String())
+	if err != nil {
+		return -1, microerror.Mask(err)
 	}
```
kopiczko, line 119: I wonder how it even worked without it:

```suggestion
	}
	defer ln.Close()
```

kopiczko (approved):

marians (approved):

pkg/callbackserver/callbackserver.go
axbarsan: lol 😬

pkg/callbackserver/callbackserver.go
```diff
@@ -72,16 +72,16 @@ func (cs *CallbackServer) Run(ctx context.Context, callback CallbackFunc) (inter
 			Handler: mux,
 		}
 	}
+	defer server.Shutdown(ctx)
 
-	// Start the server.
-	go (func(s *http.Server) {
```
axbarsan: Removed the function argument, as the scope of the goroutine will not outlive the scope of the parent function, there's no need to copy the reference

pkg/callbackserver/callbackserver.go
axbarsan: You mean having `defer server.Shutdown(ctx)`? That results in a 🔨 from the linter
Having `defer func(){ ... }` with `errorCh` also doesn't work, because at the point of `defer`, we already stopped waiting for messages on the `errorCh` channel.

pkg/callbackserver/callbackserver.go
```diff
@@ -100,7 +94,16 @@ func (cs *CallbackServer) Run(ctx context.Context, callback CallbackFunc) (inter
 		break
 	}
 
-	return result, microerror.Mask(origErr)
+	err := server.Shutdown(ctx)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	if origErr != nil {
```
axbarsan, line 102: Added this explicit error check, to not rely on `microerror.Mask()`'s underlying implementation

PR #365, opened 2021-05-27
Replace cluster scoped AppCatalog with namespace scoped Catalog CRD

rossf7:
Towards https://github.com/giantswarm/roadmap/issues/219

This updates the app commands to use the new Catalog CRD. During the migration period we will maintain both appcatalog and catalog CRs for our public catalogs.


axbarsan (approved):
LGTM

tomahawk28 (approved):
LGTM

rossf7:
We want to remove the selector flag since this is not the usual kubectl approach and can be confusing to users.

Instead we can exclude by namespace as our internal catalogs will be in the giantswarm namespace.

```
-l, --selector string                Specify label selector(s) to filter Apps by. (default "application.giantswarm.io/catalog-visibility=public")
```

cc @marians

cmd/get/catalogs/flag.go
```diff
@@ -1,24 +1,23 @@
...
-	LabelSelector string
+	AllNamespaces bool
 
 	config genericclioptions.RESTClientGetter
 	print  *genericclioptions.PrintFlags
 }
 
 func (f *flag) Init(cmd *cobra.Command) {
-	cmd.Flags().StringVarP(&f.LabelSelector, flagLabelSelector, "l", publicCatalogSelector, "Specify label selector(s) to filter Apps by.")
+	cmd.Flags().BoolVarP(&f.AllNamespaces, flagAllNamespaces, "A", false, "If present, list the requested object(s) across all namespaces. Namespace in current context is ignored even if specified with --namespace.")
```
rossf7, line 20: Label selector is replaced with all namespaces flag.

pkg/data/domain/catalog/service.go
```diff
@@ -39,68 +39,72 @@ func New(config Config) (Interface, error) {
...
 				return nil, microerror.Mask(noResourcesError)
 			}
 		}
 
-		for _, appCatalog := range appCatalogs.Items {
-			a := AppCatalog{
-				CR: omitManagedFields(appCatalog.DeepCopy()),
+		for _, catalog := range catalogs.Items {
+			// We hide catalog CRs from the giantswarm namespace by
+			// default as these are internal.
```
rossf7, line 83: If -A is true we hide internal catalogs in the giantswarm namespace.

You can still use -n giantswarm if you have access.

rossf7 (commented):
@marians Changes to remove label selector are done.

We still need to have the Catalog CRD deployed to production MCs before this can be merged.

pkg/template/catalog/catalog.go
```diff
@@ -17,48 +17,49 @@ type Config struct {
...
 			APIVersion: "application.giantswarm.io/v1alpha1",
 		},
 		ObjectMeta: metav1.ObjectMeta{
-			Name: config.Name,
-			Labels: map[string]string{
-				"app-operator.giantswarm.io/version":     "1.0.0",
-				"application.giantswarm.io/catalog-type": "awesome",
-			},
+			Name:      config.Name,
+			Namespace: config.Namespace,
```
rossf7, line 48: Dropping these labels as we don't need them.

axbarsan (approved):

PR #372, opened 2021-06-16
Add correct AWS subnet validation

pipo02mix:


anvddriesch (approved):
Not blocking here since this validation is already part of `kubectl-gs`. However, is there a reason this isn't done in the `admission-controller` instead? It tends to be better to keep all validation in one place so that we get the same result no matter how the CR was created.

pipo02mix:
Yeah I kind of agree with you but here it is a templating command and I find it annoying when you create your template successfully and then complain when the pipeline is running because of validation on the server-side. Ideally, this check is done in the server when we template command is executed it but are not there yet

anvddriesch:
Yeah, I understand. I don't know if this is even validated server-side.
(If it is not and you need it, please always drop an issue for firecracker!)

PR #377, opened 2021-06-21
Update Dockerfile

marians:


Dockerfile
```diff
@@ -1,17 +1,15 @@
...
     && mkdir -p /binaries \
     && curl -SL https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl -o /binaries/kubectl \
     && chmod +x /binaries/*
 
-FROM quay.io/giantswarm/alpine:3.12
+FROM quay.io/giantswarm/alpine:3.14
 
 COPY --from=binaries /binaries/* /usr/bin/
 COPY ./kubectl-gs /usr/bin/kubectl-gs
-RUN ln -s /usr/bin/kubectl-gs /usr/bin/kgs
```
ubergesundheit, line 14: I would say removing this symlink would break backwards compatibility. I would say we leave this in. Changing the `ENTRYPOINT` is ok though

marians: Good point, thanks

Dockerfile
```diff
@@ -1,17 +1,16 @@
-FROM quay.io/giantswarm/alpine:3.12 AS binaries
+FROM quay.io/giantswarm/alpine:3.14 AS binaries
 
-ARG KUBECTL_VERSION=1.18.9
+ARG KUBECTL_VERSION=1.29.2
```
ubergesundheit, line 3: I am wondering why this works at all.. kubectl version should be aligned with kubernetes version.

The url https://storage.googleapis.com/kubernetes-release/release/v1.29.2/bin/linux/amd64/kubectl results in a 404 respponse

Dockerfile
```diff
@@ -1,17 +1,16 @@
...
 
 RUN apk add --no-cache ca-certificates curl \
     && mkdir -p /binaries \
     && curl -SL https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl -o /binaries/kubectl \
```
ubergesundheit, line 7: add `-f` flag to make curl exit non-zero if the response was not successful

```suggestion
    && curl -SLf https://storage.googleapis.com/kubernetes-release/release/v${KUBECTL_VERSION}/bin/linux/amd64/kubectl -o /binaries/kubectl \
```

Dockerfile
marians: Oops! Good we have reviews. I read this as the kubectl-gs version.

ubergesundheit (approved):
LGTM if you tested the container image :)

PR #379, opened 2021-06-22
Extend template app to only output required fields

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/16978

Docs: https://github.com/giantswarm/docs/pull/986

The current `template app` can be improved because it defaults a lot of boiler plate fields that will be set by app-admission-controller.

We also have ugly empty fields because although fields are optional in the CRD we don't use struct pointers. So the YAML
serialization doesn't work. I'd like us to fix that in a future v1alpha2 App CRD but its non-trivial as we'll need a conversion webhook.

### Before

```yaml
go run main.go template app \
  --catalog giantswarm-playground \
  --name keda \
  --namespace default \
  --cluster 2hr7z  \
  --version 0.1.0
---
---
kind: App
metadata:
  creationTimestamp: null
  labels:
    app-operator.giantswarm.io/version: 1.0.0
  name: keda
  namespace: 2hr7z
spec:
  catalog: giantswarm-playground
  config:
    configMap:
      name: 2hr7z-cluster-values
      namespace: 2hr7z
    secret:
      name: ""
      namespace: ""
  install: {}
  kubeConfig:
    context:
      name: 2hr7z-kubeconfig
    inCluster: false
    secret:
      name: 2hr7z-kubeconfig
      namespace: 2hr7z
  name: keda
  namespace: default
  namespaceConfig: {}
  userConfig:
    configMap:
      name: ""
      namespace: ""
    secret:
      name: ""
      namespace: ""
  version: 0.1.0
status:
  appVersion: ""
  release:
    lastDeployed: null
    status: ""
  version: ""
```

### After

```yaml
go run main.go template app \
  --catalog giantswarm-playground \
  --name keda \
  --namespace default \
  --cluster 2hr7z  \
  --version 0.1.0
---
---
kind: App
metadata:
  name: keda
  namespace: 2hr7z
spec:
  catalog: giantswarm-playground
  kubeConfig:
    inCluster: false
  name: keda
  namespace: default
  version: 0.1.0
```

### DIFF
--- a/cmd/template/app/flag.go
+++ b/cmd/template/app/flag.go
@@ -6,18 +6,20 @@ import (
 )
 
 const (
-	flagCatalog       = "catalog"
-	flagCluster       = "cluster"
-	flagName          = "name"
-	flagNamespace     = "namespace"
-	flagUserConfigMap = "user-configmap"
-	flagUserSecret    = "user-secret"
-	flagVersion       = "version"
+	flagCatalog           = "catalog"
+	flagCluster           = "cluster"
+	flagDefaultingEnabled = "defaulting-enabled"
+	flagName              = "name"
+	flagNamespace         = "namespace"
+	flagUserConfigMap     = "user-configmap"
+	flagUserSecret        = "user-secret"
+	flagVersion           = "version"
 )
 
 type flag struct {
 	Catalog           string
 	Cluster           string
+	DefaultingEnabled bool
 	Name              string
 	Namespace         string
 	flagUserConfigMap string
@@ -30,6 +32,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.Name, flagName, "", "App name.")
 	cmd.Flags().StringVar(&f.Namespace, flagNamespace, "", "Namespace where the app will be deployed.")
 	cmd.Flags().StringVar(&f.Cluster, flagCluster, "", "Cluster where the app will be deployed.")
+	cmd.Flags().BoolVar(&f.DefaultingEnabled, flagDefaultingEnabled, true, "Don't template fields that will be defaulted.")
 	cmd.Flags().StringVar(&f.flagUserConfigMap, flagUserConfigMap, "", "Path to the user app configmap file data.")
 	cmd.Flags().StringVar(&f.flagUserSecret, flagUserSecret, "", "Path to the user app secret file data.")
 	cmd.Flags().StringVar(&f.Version, flagVersion, "", "App version to be installed.")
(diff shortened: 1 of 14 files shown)


cmd/template/app/flag.go
```diff
@@ -6,18 +6,20 @@ import (
...
-	flagCatalog       = "catalog"
-	flagCluster       = "cluster"
-	flagName          = "name"
-	flagNamespace     = "namespace"
-	flagUserConfigMap = "user-configmap"
-	flagUserSecret    = "user-secret"
-	flagVersion       = "version"
+	flagCatalog           = "catalog"
+	flagCluster           = "cluster"
+	flagDefaultingEnabled = "defaulting-enabled"
```
rossf7, line 11: See the docs PR but the defaulting is only enabled for these releases. As we didn't want to impact existing app CRs.

- AWS >= v14.0.0
- Azure >= v13.1.0
- KVM >= v13.1.0

If this flag is set to false then all fields will be set.

go.mod
```diff
@@ -8,7 +8,7 @@ require (
 	github.com/fatih/color v1.10.0
 	github.com/ghodss/yaml v1.0.0
 	github.com/giantswarm/apiextensions/v3 v3.22.0
-	github.com/giantswarm/app/v4 v4.12.0
+	github.com/giantswarm/app/v4 v4.11.0
```
rossf7, line 11: I think I broke this as I had to downgrade or I got compile errors.

There is nothing relevant we need in the 4.12.0 release.

pkg/template/app/app.go
```diff
@@ -120,3 +130,64 @@ func NewSecret(config SecretConfig) (*corev1.Secret, error) {
 
 	return secret, nil
 }
+
+// printAppCR removes empty fields from the app CR YAML. This is needed because
+// although the fields are optional we do not use struct pointers. This will
+// be fixed in a future version of the App CRD.
```
rossf7, line 136: This is nasty but I don't see another way of stripping the empty fields. 

So I've tried to make sure there is good test coverage.

axbarsan: Does the YAML issue still exist if using `sigs.k8s.io/yaml`?

https://github.com/giantswarm/fmt/blob/master/go/yaml_marshalling.md

rossf7: @axbarsan No sadly the empty fields are still shown :(

The problem is although we have JSON tags with`omitempty` we use regular structs not struct pointers.

Theo tried to introduce this in https://github.com/giantswarm/apiextensions/pull/283. It works well but the problem is we need to handle the nil structs and the code changes got big.

The safest way I can see to introduce these changes is with a new v1alpha2 version of the App CR.

pipo02mix (approved):
that is noice! happy when Adrian is happy

axbarsan (approved):
LGTM

pkg/template/app/app.go
rossf7: @axbarsan I've switched to use `sigs.k8s.io/yaml` to match fmt. We were mostly using that anyway.

I've created https://github.com/giantswarm/roadmap/issues/365 for making the fields optional in the App CRD by using struct pointers.

We already did this for the new Catalog CRD that replaces AppCatalog.

PR #380, opened 2021-06-28
capa cluster templating

anvddriesch:
We now have giant swarm release `v.20.0.0` for aws that uses the upstream CAPI and CAPA controllers. This version is a test version that is currently used internally. This PR adds templating for Clusters and Machinepools in this version to `kubectl-gs` so that our teams can create test clusters more easily.
The templates are directly pulled from upstream but we add our labels, cross account management and (most of) the options currently supported in `kubectl-gs`. The templates need to still be defauled by `kyverno` when they are applied.

Towards https://github.com/giantswarm/giantswarm/issues/17831


calvix (approved):

cmd/template/nodepool/provider/aws.go
```diff
@@ -64,3 +149,59 @@ func WriteAWSTemplate(out io.Writer, config NodePoolCRsConfig) error {
 
 	return nil
 }
+
+func getCAPANodepoolTemplate(config NodePoolCRsConfig) (client.Template, error) {
```
anvddriesch, line 153: This function uses the upstream client to fetch the CAPA templates directly from the repo the way `clusterctl` does. The idea would be to extend this also to the other providers and versions than just `aws v20.0.0`

cmd/template/nodepool/provider/aws.go
```diff
@@ -64,3 +149,59 @@ func WriteAWSTemplate(out io.Writer, config NodePoolCRsConfig) error {
...
+
+	c, err := client.New("")
+	if err != nil {
+		return nil, err
+	}
+
+	templateOptions := client.GetClusterTemplateOptions{
+		ClusterName:       config.ClusterID,
+		TargetNamespace:   config.Owner,
+		KubernetesVersion: "v1.19.9",
```
anvddriesch, line 164: We need to find a way to get the desired kubernetes version here. If it should come from the release CR, we need to default it later. However, the request will not get through when this here is not set. We can also consider getting it from a flag, but that would open up customers to use potentially untested versions.

cmd/template/nodepool/provider/aws.go
```diff
@@ -64,3 +149,59 @@ func WriteAWSTemplate(out io.Writer, config NodePoolCRsConfig) error {
...
+		ProviderRepositorySource: &client.ProviderRepositorySourceOptions{
+			InfrastructureProvider: "aws:v0.6.6",
+			Flavor:                 "machinepool",
+		},
+	}
+	os.Setenv("AWS_SUBNET", "")
+	os.Setenv("AWS_CONTROL_PLANE_MACHINE_TYPE", "")
+	os.Setenv("AWS_REGION", "")
+	os.Setenv("AWS_SSH_KEY_NAME", "")
```
anvddriesch, line 173: For some reason the client DEMANDS these envs to be set. Left empty for defaulting.

internal/key/key.go
```diff
@@ -58,6 +58,12 @@ func GenerateAssetName(values ...string) string {
 	return strings.Join(values, "-")
 }
 
+// IsCAPAVersion returns whether a given GS Release Version is based on the CAPI/CAPA projects
+// TODO: make this a >= comparison
+func IsCAPAVersion(version string) bool {
+	return version == "20.0.0"
```
anvddriesch, line 64: Once again, this should be opened up eventually.

cmd/template/nodepool/runner.go
```diff
@@ -59,6 +59,7 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 			OnDemandBaseCapacity:                r.flag.OnDemandBaseCapacity,
 			OnDemandPercentageAboveBaseCapacity: r.flag.OnDemandPercentageAboveBaseCapacity,
 			Owner:                               r.flag.Owner,
+			ReleaseVersion:                      r.flag.Release,
```
anvddriesch, line 62: We have to add a release flag to the nodepool command in order to recognize when we want a CAPA machinepool.

cmd/template/cluster/provider/aws.go
```diff
@@ -76,3 +171,51 @@ func WriteAWSTemplate(out io.Writer, config ClusterCRsConfig) error {
...
+func newAWSMachineTemplateFromUnstructured(config ClusterCRsConfig, o unstructured.Unstructured) (*capav1alpha3.AWSMachineTemplate, error) {
+	var awsmachinetemplate capav1alpha3.AWSMachineTemplate
+	{
+		err := runtime.DefaultUnstructuredConverter.
+			FromUnstructured(o.Object, &awsmachinetemplate)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+		if config.ControlPlaneAZ != nil {
+			awsmachinetemplate.Spec.Template.Spec.FailureDomain = &config.ControlPlaneAZ[0]
```
anvddriesch, line 217: This is unfortunate but it seems like we can only have **one** AZ for the control-plane, even if several masters are running there. :( Not sure what is the most elegant way to deal with it.

njuettner: Hmm maybe we could try something else. If I got it right from the docs, you can specify `networkSpec` in order to select your AZ's for control-plane, if you don't want to have random AZ's, https://cluster-api-aws.sigs.k8s.io/topics/multi-az-control-planes.html#overview. This is set in the AWSCluster.

With this approach we would basically force all master nodes to be in the same region, not sure if we want that as you said 🤔

cmd/template/nodepool/provider/aws.go
njuettner: What is the reason using this approach vs. getting the templates from apiextensions? Is this more convenient?

anvddriesch: Our idea was to stay close to the upstream and always get the right templates for the controller version and flavor we want to use. For now it seems convenient since we are still learning. Maybe in the future we want to have our own default templates.

cmd/template/cluster/provider/aws.go
anvddriesch: Oh nice, I did not see that! :) I'll try to use this.

anvddriesch:
:) Pinged some more people for feedback. If it's okay with everyone I would like to merge this so we can create clusters and machinepools with v20.0.0 more easily.

cmd/template/cluster/provider/aws.go
```diff
@@ -2,19 +2,139 @@ package provider
...
+		AWSMachineTemplateCR     string
+		ClusterCR                string
+		KubeadmControlPlaneCR    string
+		AWSClusterRoleIdentityCR string
+	}{}
+
+	crLabels := map[string]string{
+		label.ReleaseVersion:            config.ReleaseVersion,
+		label.Cluster:                   config.ClusterID,
+		"cluster.x-k8s.io":              config.ClusterID,
```
axbarsan, line 66: On the Azure implementation we use `cluster.x-k8s.io/cluster-name`.
https://github.com/giantswarm/kubectl-gs/blob/4027733d3b3954de75f8535e19b3bb32c11f810e/cmd/template/cluster/provider/azure.go#L81

Is the different label used on purpose?

cmd/template/cluster/provider/aws.go
```diff
@@ -76,3 +196,97 @@ func WriteAWSTemplate(out io.Writer, config ClusterCRsConfig) error {
...
+			Name: config.ClusterID,
+			Kind: capav1alpha3.ClusterRoleIdentityKind}
+
+		if config.ControlPlaneAZ != nil {
+			for _, az := range config.ControlPlaneAZ {
+				privateSubnet := capav1alpha3.SubnetSpec{AvailabilityZone: az, IsPublic: false}
+				publicSubnet := capav1alpha3.SubnetSpec{AvailabilityZone: az, IsPublic: true}
+				awscluster.Spec.NetworkSpec.Subnets = append(awscluster.Spec.NetworkSpec.Subnets, &privateSubnet, &publicSubnet)
+			}
+		}
```
axbarsan, line 251: There's no need for the `nil` check. `for ... range ...` already takes that into account

```suggestion
		for _, az := range config.ControlPlaneAZ {
			privateSubnet := capav1alpha3.SubnetSpec{AvailabilityZone: az, IsPublic: false}
			publicSubnet := capav1alpha3.SubnetSpec{AvailabilityZone: az, IsPublic: true}
			awscluster.Spec.NetworkSpec.Subnets = append(awscluster.Spec.NetworkSpec.Subnets, &privateSubnet, &publicSubnet)
		}
```

marians:
@anvddriesch Would you mind adding some explanation to the PR? I don't know on what base to review, so something about goal/purpose/scope would be great. Also the repo is public, so some explanation would help users understand changes (in retrospect mainly).

cmd/template/nodepool/provider/aws.go
```diff
@@ -15,6 +23,92 @@ import (
...
+	}{}
+
+	objects := nodepoolTemplate.Objs()
+	for _, o := range objects {
+		o.SetName(config.NodePoolID)
+		o.SetLabels(map[string]string{
+			label.ReleaseVersion:            config.ReleaseVersion,
+			label.Cluster:                   config.ClusterID,
+			label.MachinePool:               config.NodePoolID,
+			"cluster.x-k8s.io":              config.ClusterID,
```
axbarsan, line 66: Same question here, with the label

cmd/template/nodepool/provider/aws.go
axbarsan: Can't we set it in the admission controller, based on the release version?

axbarsan (commented):
LGTM in general, just a couple questions

cmd/template/cluster/provider/aws.go
anvddriesch: I'll check but I think this is just a typo. good catch.

cmd/template/nodepool/provider/aws.go
anvddriesch: Unfortunately not. If this is not set in the `templateOptions` the request is blocked and we do not get a template that could be defaulted by the admission-controller.

anvddriesch:
@marians I have written a small summary in the PR description. I hope that helps. I will also add the CHANGELOG.

cmd/template/nodepool/provider/aws.go
ubergesundheit: I think we should check if these are already set before setting them to an empty string and also [unset](https://golang.org/pkg/os/#Unsetenv) these once no longer required.

marians:
Thanks for adding the PR info!

### User guidance

**So how do I create a CAPA cluster manifest?** I had to think hard to come to the conclusion that when adding `--release 20.0.0` to the `template cluster`, I'll probably get a CAPA cluster manifest. I wonder whether we have to make that more discoverable, either in the command's usage info, or in documentation. (I'd say we SHOULD mention it in docs and MAY mention it in the usage info).

### Regarding `template cluster` output

In the `KubeadmControlPlane` I see two strings that look like templates instead of values. Is that as it should be? Snippet:

```yaml
spec:
  kubeadmConfigSpec:
    initConfiguration:
      nodeRegistration:
        name: '{{ ds.meta_data.local_hostname }}'
    joinConfiguration:
      nodeRegistration:
        name: '{{ ds.meta_data.local_hostname }}'
```

### Regarding `template nodepool` output

Same question as above, is this fine in the `KubeadmConfig`?

```yaml
spec:
  joinConfiguration:
    nodeRegistration:
      name: '{{ ds.meta_data.local_hostname }}'
```

cmd/template/nodepool/provider/aws.go
calvix: Yeah, +1 on that

marians:
I just noticed that `template nodepool` for v20 creates `.status` sections in the output (in the MachinePool resource). That's a bit strange to me. Can that be avoided?

njuettner (approved):

axbarsan (approved):

marians:
I just learned from @calvix that our goal is not to make this discoverable by non Giant Swarm users, as it's only functional in test installations currently.

marians (approved):

cmd/template/nodepool/provider/aws.go
anvddriesch: I agree with unsetting them! Will add that.
However, if we check if they are already set and leave that value, it could lead to unexpected different outcomes. I would prefer to leave them as empty strings and transparently default in kyverno.

anvddriesch:
The `spec` values of `KubeadmControlPlane` and `KubeadmConfig` are filled in this way based on the upstream templating. I would leave it like this for now until we are more familiar with the creation of CAPI clusters and can add our own templates.
The same is true for the Machinepool status, but I agree that it is odd. I'll try what happens if we remove the status at least.

And yeah, the goal for now is just for us in giantswarm to be able to create clusters and experiment with v20.0.0.

cmd/template/nodepool/provider/aws.go
ubergesundheit: makes sense I guess

ubergesundheit: Was this the right way to solve this though? In environments having these set, we should revert them to the value before, so something like

```go
if os.Getenv("AWS_SUBNET") != "" {
  prevEnvAWSSubnet := os.Getenv("AWS_SUBNET")
  defer os.Setenv("AWS_SUBNET", prevEnvAWSSubnet)
} else {
  defer os.Unsetenv("AWS_SUBNET")
}
```

anvddriesch: True, good point. Follow up: https://github.com/giantswarm/kubectl-gs/pull/390

PR #383, opened 2021-07-07
Allow overriding the app name in template app

rossf7:
See https://github.com/giantswarm/docs/pull/912#issuecomment-875634640

Currently the `--name` param in `template app` is used for both the name of the app in the catalog and the app CR name. However in some situations like installing multiple ingress controllers you also need to be able to set the app name.

This adds a new optional `--app-name` param for this.


axbarsan (approved):
Nice

PR #391, opened 2021-07-13
trim v in release version for nodepools

anvddriesch:
We had the problem that using the option `--release v20.0.0` and `--release 20.0.0` would both result in a CAPI cluster being templated but in case of nodepools, it had to be `--release 20.0.0`. This PR makes sure that the `v` is trimmed in nodepool templating as well and both versions work.

### DIFF
--- a/cmd/template/nodepool/runner.go
+++ b/cmd/template/nodepool/runner.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"io"
 	"os"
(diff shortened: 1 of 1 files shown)


calvix (approved):
i assume we do the same for cluster so this is fine

anvddriesch:
Yes, this is what I copy pasted from the template cluster command. :)

PR #394, opened 2021-07-15
Extend login error message to mention OIDC config

rossf7:
@cokiengchiara Suggested adding this after the problem with bandicoot yesterday.

I think this is a pretty common case and so hopefully the error message improves things. WDYT?

### DIFF
--- a/pkg/installation/info.go
+++ b/pkg/installation/info.go
@@ -43,7 +43,7 @@ func getInstallationInfo(ctx context.Context, gqlClient graphql.Client) (install
 	var info installationInfo
 	err := gqlClient.ExecuteQuery(ctx, infoQuery, nil, &info)
 	if err != nil {
-		return installationInfo{}, microerror.Maskf(cannotGetInstallationInfoError, "make sure you're connected to the internet and that the Athena service is up and running\n%s", err.Error())
+		return installationInfo{}, microerror.Maskf(cannotGetInstallationInfoError, "make sure you're connected to the internet, OIDC is configured for the management cluster and that the Athena service is up and running\n%s", err.Error())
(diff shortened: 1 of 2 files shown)


axbarsan (approved):
Nice 👍

pkg/installation/info.go
```diff
@@ -43,7 +43,7 @@ func getInstallationInfo(ctx context.Context, gqlClient graphql.Client) (install
 	var info installationInfo
 	err := gqlClient.ExecuteQuery(ctx, infoQuery, nil, &info)
 	if err != nil {
-		return installationInfo{}, microerror.Maskf(cannotGetInstallationInfoError, "make sure you're connected to the internet and that the Athena service is up and running\n%s", err.Error())
+		return installationInfo{}, microerror.Maskf(cannotGetInstallationInfoError, "make sure you're connected to the internet, OIDC is configured for the management cluster and that the Athena service is up and running\n%s", err.Error())
```
marians, line 46:
```suggestion
		return installationInfo{}, microerror.Maskf(cannotGetInstallationInfoError, "make sure you're connected to the internet, OIDC authentication is configured for the Management API, and the Athena service is up and running\n%s", err.Error())
```

My suggestion is about adding `authentication` and replacing management cluster with `Management API`.

If that oxford comma doesn't make sense here, please fix as you think.

cokiengchiara (approved):
LGTM, thank you! I like Marian's suggestion.

pkg/installation/info.go
rossf7: Thanks Marian much better and good to add the oxford comma. 👍

PR #397, opened 2021-07-19
Add support for spot VMs on azure machine pool template cmd

whites11:
Towards: https://github.com/giantswarm/roadmap/issues/372

This PR allows creating node pools with azure spot VMs

### DIFF
--- a/cmd/template/nodepool/flag.go
+++ b/cmd/template/nodepool/flag.go
@@ -87,7 +87,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 	// Azure only.
 	cmd.Flags().StringVar(&f.AzureVMSize, flagAzureVMSize, "Standard_D4s_v3", "Azure VM size to use for workers, e.g. 'Standard_D4s_v3'.")
 	cmd.Flags().BoolVar(&f.AzureUseSpotVms, flagAzureUseSpotVMs, false, "Whether to use Spot VMs for this Node Pool. Defaults to false.")
-	cmd.Flags().Float32Var(&f.AzureSpotMaxPrice, flagAzureSpotMaxPrice, 0, "Max hourly price in USD to pay for one spot VM. If not set than on demand pricing is used as the max value.")
+	cmd.Flags().Float32Var(&f.AzureSpotVMsMaxPrice, flagAzureSpotVMsMaxPrice, 0, "Max hourly price in USD to pay for one spot VM on Azure. If not set, the on-demand price is used as the limit.")
 
 	// Common.
 	cmd.Flags().StringSliceVar(&f.AvailabilityZones, flagAvailabilityZones, []string{}, "List of availability zones to use, instead of setting a number. Use comma to separate values.")
--- a/cmd/template/nodepool/flag.go
+++ b/cmd/template/nodepool/flag.go
@@ -21,7 +21,9 @@ const (
 	flagUseAlikeInstanceTypes               = "use-alike-instance-types"
 
 	// Azure only.
-	flagAzureVMSize = "azure-vm-size"
+	flagAzureVMSize       = "azure-vm-size"
+	flagAzureUseSpotVMs   = "azure-spot-vms"
+	flagAzureSpotMaxPrice = "azure-spot-max-price"
 
 	// Common.
 	flagAvailabilityZones = "availability-zones"
@@ -52,7 +54,9 @@ type flag struct {
 	UseAlikeInstanceTypes               bool
 
 	// Azure only.
-	AzureVMSize string
+	AzureVMSize       string
+	AzureUseSpotVms   bool
+	AzureSpotMaxPrice float32
 
 	// Common.
 	AvailabilityZones []string
@@ -82,6 +86,8 @@ func (f *flag) Init(cmd *cobra.Command) {
 
 	// Azure only.
 	cmd.Flags().StringVar(&f.AzureVMSize, flagAzureVMSize, "Standard_D4s_v3", "Azure VM size to use for workers, e.g. 'Standard_D4s_v3'.")
+	cmd.Flags().BoolVar(&f.AzureUseSpotVms, flagAzureUseSpotVMs, false, "Whether to use Spot VMs for this Node Pool.")
+	cmd.Flags().Float32Var(&f.AzureSpotMaxPrice, flagAzureSpotMaxPrice, 0, "Max hourly price in USD to pay for one spot VM.")
 
 	// Common.
 	cmd.Flags().StringSliceVar(&f.AvailabilityZones, flagAvailabilityZones, []string{}, "List of availability zones to use, instead of setting a number. Use comma to separate values.")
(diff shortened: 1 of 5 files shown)


fiunchinho (approved):
LGTM

cmd/template/nodepool/provider/azure.go
```diff
@@ -73,6 +75,18 @@ func WriteAzureTemplate(out io.Writer, config NodePoolCRsConfig) error {
...
 func newAzureMachinePoolCR(config NodePoolCRsConfig) *expcapzv1alpha3.AzureMachinePool {
+	var spot *v1alpha3.SpotVMOptions
+	if config.AzureUseSpotVms {
+		var maxPrice *resource.Quantity
+		if config.AzureSpotMaxPrice > 0 {
+			maxPriceScalar := resource.MustParse(fmt.Sprintf("%f", config.AzureSpotMaxPrice))
+			maxPrice = &maxPriceScalar
+		}
+		spot = &v1alpha3.SpotVMOptions{
+			MaxPrice: maxPrice,
```
axbarsan, line 86: Is leaving this as `nil` the same as setting it to `-1` (using on-demand pricing as max)?

axbarsan (commented):
Please also update the docs
https://docs.giantswarm.io/ui-api/kubectl-gs/template-nodepool/

cmd/template/nodepool/provider/azure.go
whites11: yeah. When setting `-1` on happa, it becomes a `nil` in the AzureMachinePool CR

axbarsan: Not currently
https://github.com/giantswarm/api/blob/2125e0f95ea111fe9c43cdd477d9f68d32959338/service/nodepool/creator/service.go#L661

whites11: Fuck you are right:

https://docs.microsoft.com/en-us/azure/templates/microsoft.compute/virtualmachinescalesets?tabs=json#BillingProfile

-1 is the needed value

whites11: done

cmd/template/nodepool/flag.go
```diff
@@ -82,6 +86,8 @@ func (f *flag) Init(cmd *cobra.Command) {
 
 	// Azure only.
 	cmd.Flags().StringVar(&f.AzureVMSize, flagAzureVMSize, "Standard_D4s_v3", "Azure VM size to use for workers, e.g. 'Standard_D4s_v3'.")
+	cmd.Flags().BoolVar(&f.AzureUseSpotVms, flagAzureUseSpotVMs, false, "Whether to use Spot VMs for this Node Pool.")
+	cmd.Flags().Float32Var(&f.AzureSpotMaxPrice, flagAzureSpotMaxPrice, 0, "Max hourly price in USD to pay for one spot VM.")
```
axbarsan, line 90: Should we default this to `-1`, and specify that by default it uses on-demand pricing as max?

whites11: The default is really not important as it gets set to -1 if it's not > 0 in modepool/provider/azure.go
Adjusted description

marians (changes requested):
Longer description coming...

marians:
Here is the current help output with this branch:

```nohighlight
Template Giant Swarm Node Pool CRs.

Usage:
  kubectl gs template nodepool [flags]

Flags:
      --availability-zones strings                     List of availability zones to use, instead of setting a number. Use comma to separate values.
      --aws-instance-type string                       EC2 instance type to use for workers, e. g. 'm5.2xlarge'. (default "m5.xlarge")
      --azure-spot-max-price float32                   Max hourly price in USD to pay for one spot VM. If not set than on demand pricing is used as the max value.
      --azure-spot-vms                                 Whether to use Spot VMs for this Node Pool. Defaults to false.
      --azure-vm-size string                           Azure VM size to use for workers, e.g. 'Standard_D4s_v3'. (default "Standard_D4s_v3")
...
      --on-demand-base-capacity int                    Number of base capacity for On demand instance distribution. Default is 0.
      --on-demand-percentage-above-base-capacity int   Percentage above base capacity for On demand instance distribution. Default is 100. (default 100)
      --use-alike-instance-types                       Whether to use similar instances types as a fallback.
...
```

We have several flags that are only usable for Spot instances on EC2 and the new flags for Azure. We should make them easy to distinguish, so that Azure users are 100% clear which ones to uses, and AWS users likewise.

(Unfortunately, we haven't prefixed the AWS only flags with `aws`. But let's fix that in another project.)

For the Azure-only flags we should add a phrase like "on Azure" to the description. Likewise, for the AWS only flags, we should add "on AWS". I'll add suggestions for the flags added in this PR, however I can't do that for the existing flags I think.

In addition, I suggest to rename the flag `--azure-spot-max-price` to `--azure-spot-vms-max-price`, as the feature is named "spot VMs", not "spot".

cmd/template/nodepool/flag.go
```diff
@@ -21,7 +21,9 @@ const (
 	flagUseAlikeInstanceTypes               = "use-alike-instance-types"
 
 	// Azure only.
-	flagAzureVMSize = "azure-vm-size"
+	flagAzureVMSize       = "azure-vm-size"
+	flagAzureUseSpotVMs   = "azure-spot-vms"
+	flagAzureSpotMaxPrice = "azure-spot-max-price"
```
marians, line 26:
```suggestion
	flagAzureSpotMaxPrice = "azure-spot-vms-max-price"
```

cmd/template/nodepool/flag.go
```diff
@@ -52,7 +54,9 @@ type flag struct {
 	UseAlikeInstanceTypes               bool
 
 	// Azure only.
-	AzureVMSize string
+	AzureVMSize       string
+	AzureUseSpotVms   bool
+	AzureSpotMaxPrice float32
```
marians, line 59:
```suggestion
	AzureSpotVMsMaxPrice float32
```

cmd/template/nodepool/flag.go
```diff
@@ -82,6 +86,8 @@ func (f *flag) Init(cmd *cobra.Command) {
 
 	// Azure only.
 	cmd.Flags().StringVar(&f.AzureVMSize, flagAzureVMSize, "Standard_D4s_v3", "Azure VM size to use for workers, e.g. 'Standard_D4s_v3'.")
+	cmd.Flags().BoolVar(&f.AzureUseSpotVms, flagAzureUseSpotVMs, false, "Whether to use Spot VMs for this Node Pool. Defaults to false.")
+	cmd.Flags().Float32Var(&f.AzureSpotMaxPrice, flagAzureSpotMaxPrice, 0, "Max hourly price in USD to pay for one spot VM. If not set than on demand pricing is used as the max value.")
```
marians, line 90:
```suggestion
	cmd.Flags().Float32Var(&f.AzureSpotVMsMaxPrice, flagAzureSpotVMsMaxPrice, 0, "Max hourly price in USD to pay for one spot VM. If not set, the on-demand price is used as the max value.")
```

cmd/template/nodepool/flag.go
```diff
@@ -82,6 +86,8 @@ func (f *flag) Init(cmd *cobra.Command) {
 
 	// Azure only.
 	cmd.Flags().StringVar(&f.AzureVMSize, flagAzureVMSize, "Standard_D4s_v3", "Azure VM size to use for workers, e.g. 'Standard_D4s_v3'.")
+	cmd.Flags().BoolVar(&f.AzureUseSpotVms, flagAzureUseSpotVMs, false, "Whether to use Spot VMs for this Node Pool. Defaults to false.")
```
marians, line 89:
```suggestion
	cmd.Flags().BoolVar(&f.AzureUseSpotVms, flagAzureUseSpotVMs, false, "Whether to use Spot VMs for this Node Pool. Defaults to false. Only available on Azure.")
```

whites11:
makes sense, thanks @marians

marians (approved):
Unblocking. Would be great to still adjust the AWS-only flag descriptions, to not let users wonder whether they apply to Azure, too.

whites11:
Added " Only available on AWS." Merging. Thanks

PR #401, opened 2021-07-23
aws-capa-add-bastion-crs

calvix:
* generate CRs and resources to create a usable bastion host in CAPA
* configure machines to be accessible via SSH
the bastion machine runs flatcar and is configured via ignition defined in the bootstrap secret
workers and cp run ubuntu for now so they need a little bit different sshd configurations


calvix:
not the best solution put all of this into `kubectl-gs` but I have no better idea atm,  there might be changes in future since with flatcar the definition for machines might look different, and also we need to consider its still alpha3/alpha4 we still have alpha5 and maybe than its stable

calvix:
```
 #: opsctl ssh gauss 54.74.17.126
Flatcar Container Linux by Kinvolk stable 2905.2.0 for Amazon EC2
giantswarm@s4w0h-bastion ~ $ ssh  10.0.114.4
The authenticity of host '10.0.114.4 (10.0.114.4)' can't be established.
ED25519 key fingerprint is SHA256:Mv4L9eY3J9iMr/kmAqi/oJVjQZJDIOGI9V5rwNjI2Jc.
This key is not known by any other names
Are you sure you want to continue connecting (yes/no/[fingerprint])? yes
Warning: Permanently added '10.0.114.4' (ED25519) to the list of known hosts.
Welcome to Ubuntu 18.04.5 LTS (GNU/Linux 5.4.0-1041-aws x86_64)
...
Ubuntu comes with ABSOLUTELY NO WARRANTY, to the extent permitted by
applicable law.

To run a command as administrator (user "root"), use "sudo <command>".
See "man sudo_root" for details.

giantswarm@ip-10-0-114-4:~$ 
```

whites11:
Wow, this is much more code than I thought would have been needed for this feature.
I only have one concern. Would it be hard to make the number of bastions and the availability zones they are in configurable?

anvddriesch (commented):
It looks good from what I can tell! However, I am really not sure if `kubectl-gs` is the right place for it. Is there any alternative?

calvix:
> Wow, this is much more code than I thought would have been needed for this feature.
> I only have one concern. Would it be hard to make the number of bastions and the availability zones they are in configurable?

the thing is that there is code for bastion and then code for worker and master nodes as well
Creating CRs takes a lot lines of code

calvix:
> It looks good from what I can tell! However, I am really not sure if `kubectl-gs` is the right place for it. Is there any alternative?

I had conversation with Marcel and Christian and we decided that this is the "best" place for it now as we don't want admission controller to create CRs

anvddriesch:
> I had conversation with Marcel and Christian and we decided that this is the "best" place for it now as we don't want admission controller to create CRs

Okay, sounds fair. Maybe get one UX opinion on it before merging,

MarcelMue (approved):
We discussed this - for the first iteration it's okay with me :)

calvix:
Hello @giantswarm/sig-ux  this does not do anything extra from UX side ( no new options args etc) It just add some Crs and adjust values of some CRs to have bastions and ssh access to nodes.

marians:
A CHANGELOG entry in common language would be appreciated 💐

axbarsan (commented):
So now we have API calls in the `template *` commands? This is something we wanted to avoid until now.
If there are API calls and K8s API access involved, the scope of the command is much bigger than just "templating", which is what the command name says. This is confusing to users.

Is it possible to avoid that? To me it looks like setting the properties that require API calls should be moved into a mutation webhook.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
+- Add CRs to create bastion host, when creating a CAPI cluster.
+- Add configuration to allow SSH for giantswarm employee when creating CAPI cluster.
```
whites11, line 13:
```suggestion
- Add configuration to allow SSH for Giant Swarm employees when creating CAPI cluster.
```

calvix:
> So now we have API calls in the `template *` commands? This is something we wanted to avoid until now.
> If there are API calls and K8s API access involved, the scope of the command is much bigger than just "templating", which is what the command name says. This is confusing to users.
> 
> Is it possible to avoid that? To me it looks like setting the properties that require API calls should be moved into a mutation webhook.

 
so we already investigated that path before this and its just not possible ATM

`clusterctl`  is already using API call to fetch CRDs from CAPI MC, that is already there before me adding another API call, its just how clustectl logic works and we use clusterctl libs to do the same

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
+- Add CRs to create bastion host, when creating a CAPI cluster.
+- Add configuration to allow SSH for Giant Swarm employees when creating CAPI cluster.
```
marians, line 13: I'm missing information regarding which command is affected and how.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
+- Add CRs to create bastion host, when creating a CAPI cluster via `template cluster` command.
+- Add configuration to allow SSH for Giant Swarm employees when creating CAPI cluster. Applies to `template cluster` and `template nodepool` commands.
```
calvix, line 13: @marians something like this is enought?

axbarsan (approved):

CHANGELOG.md
marians: Superb, thanks!

marians (approved):

PR #426, opened 2021-08-18
Update kubectl and base image in Dockerfile

marians:
This PR updates kubectl from v1.18 to v1.19 and also updates the alpine base image used.


ubergesundheit (approved):
Generally okay. I assume you have a good reason to pin to this specific alpine version. This has to do with the way we retag the container images? `3.14` does not point to `3.14.1` right?

marians:
Yes, 3.14 I think is 3.14.0. We don't move that tag to newer patch versions. 3.14.1 is our newest alpine.

PR #436, opened 2021-08-23
Make the `login` command work even if opening the browser fails

axbarsan:
The `login` command fails if opening the default browser fails. With this PR, if that happens, we print a message to the console and we keep the auth proxy running, so the user could open the URL by themselves and continue the authentication process.

### DIFF
--- a/cmd/login/auth.go
+++ b/cmd/login/auth.go
@@ -40,7 +40,7 @@ var (
 )
 
 // handleAuth executes the OIDC authentication against an installation's authentication provider.
-func handleAuth(ctx context.Context, out io.Writer, i *installation.Installation, clusterAdmin bool) (oidc.UserInfo, error) {
+func handleAuth(ctx context.Context, out io.Writer, errOut io.Writer, i *installation.Installation, clusterAdmin bool) (oidc.UserInfo, error) {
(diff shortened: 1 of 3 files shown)


marians (approved):

cmd/login/auth.go
```diff
@@ -76,19 +76,19 @@ func handleAuth(ctx context.Context, out io.Writer, i *installation.Installation
...
-	}
 
 	fmt.Fprintf(out, "\n%s\n", color.YellowString("Your browser should now be opening this URL:"))
 	fmt.Fprintf(out, "%s\n\n", authURL)
 
+	// Open the authorization url in the user's browser, which will eventually
+	// redirect the user to the local web server we'll create next.
+	err = open.Run(authURL)
+	if err != nil {
+		fmt.Fprintf(errOut, "%s\n\n", color.RedString("Couldn't open the default browser. Please access the URL above to continue logging in."))
```
kopiczko, line 89: Feels like it should be yellow as you can recover.

kopiczko (approved):

cmd/login/auth.go
marians: I don't think we have a strong definition regarding what's allowed to go to stderr. In this case I think either stdout or stderr is fine.

PR #453, opened 2021-09-03
- Added support to generate templates for CAPZ node pools.

whites11:
Towards: https://github.com/giantswarm/giantswarm/issues/18620

Added support to generate capz node pool templates

### DIFF
--- a/internal/key/key.go
+++ b/internal/key/key.go
@@ -220,3 +220,30 @@ func OrganizationNamespaceFromName(name string) string {
 
 	return name
 }
+
+func AzureStorageAccountTypeForVMSize(vmSize string) string {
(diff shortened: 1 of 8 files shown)


internal/key/key.go
```diff
@@ -220,3 +220,30 @@ func OrganizationNamespaceFromName(name string) string {
 
 	return name
 }
+
+func AzureStorageAccountTypeForVMSize(vmSize string) string {
+	// Ugly but hardcoding is the only way to know premium storage support without reaching Azure API.
+	vmTypesWithPremiumStorageSupport := map[string]bool{
```
MarcelMue, line 226: Yeah this is not very fun :(

MarcelMue (approved):

PR #461, opened 2021-09-08
Allow providing Kubernetes API URLs with prefixes to the login command.

axbarsan:
Depends on https://github.com/giantswarm/config/pull/434

This makes the `login` command smarter by allowing the usage of Kubernetes API URLs with prefixes. This fixes login on the `viking` installation.

### DIFF
--- a/pkg/installation/url.go
+++ b/pkg/installation/url.go
@@ -21,6 +21,8 @@ const (
 	UrlTypeHappa
 )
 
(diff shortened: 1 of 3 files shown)


pkg/installation/url.go
```diff
@@ -21,6 +21,8 @@ const (
 	UrlTypeHappa
 )
 
+var k8sApiURLRegexp = regexp.MustCompile(fmt.Sprintf("([^.]*)%s.*$", k8sApiUrlPrefix))
```
axbarsan, line 24: Kubernetes API URLs no longer need to start with `g8s.`

kopiczko (approved):

PR #467, opened 2021-09-10
Start a new auth flow if the refresh token or the id token are missing

axbarsan:
Closes https://github.com/giantswarm/giantswarm/issues/18285

Fairly straightforward. If the refresh token or ID token are missing, we execute the authentication flow again, since we still have the K8s API URL and the installation codename stored.


marians (approved):
Thanks!

PR #463, opened 2021-09-09
Default using webhooks from managementcluster

MarcelMue:
When creating a pull request for `kubectl-gs`, please consider:

- Provide a link to the issue, and please describe the goal you are trying to accomplish in this description.
- SIG UX cares about almost all changes here and is always happy to offer feedback. Simply request a review.
- Add a user-friendly description of your change to `CHANGELOG.md`.
- Our public [kubectl-gs documentation](https://docs.giantswarm.io/ui-api/kubectl-gs/) may need an update to reflect the change you are making.

PR #471, opened 2021-09-13
Add organization flag and deprecate owner instead

pipo02mix:
There is no official issue but SIG UX discussion and make a lot of sense to align the naming here https://gigantic.slack.com/archives/C13MG9F0W/p1631274549023400

### DIFF
--- a/cmd/template/cluster/flag.go
+++ b/cmd/template/cluster/flag.go
@@ -26,7 +26,7 @@ const (
 	flagMasterAZ            = "master-az" // TODO: Remove some time after August 2021
 	flagName                = "name"
 	flagOutput              = "output"
-	flagOrganization        = "org"
+	flagOrganization        = "organization"
 	flagOwner               = "owner" // TODO: Remove some time after December 2021
 	flagRelease             = "release"
 	flagLabel               = "label"
--- a/cmd/template/cluster/flag.go
+++ b/cmd/template/cluster/flag.go
@@ -26,7 +26,8 @@ const (
 	flagMasterAZ            = "master-az" // TODO: Remove some time after August 2021
 	flagName                = "name"
 	flagOutput              = "output"
-	flagOwner               = "owner"
+	flagOrganization        = "org"
+	flagOwner               = "owner" // TODO: Remove some time after December 2021
 	flagRelease             = "release"
 	flagLabel               = "label"
 )
@@ -46,6 +47,7 @@ type flag struct {
 	MasterAZ            []string
 	Name                string
 	Output              string
+	Organization        string
 	Owner               string
 	Release             string
 	Label               []string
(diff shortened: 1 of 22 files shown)


cmd/template/cluster/flag.go
```diff
@@ -129,8 +133,6 @@ func (f *flag) Validate() error {
 				return microerror.Maskf(invalidFlagError, "--%s must be a valid subnet size (20, 21, 22, 23, 24 or 25)", flagControlPlaneSubnet)
 			}
 		}
-
-		return nil
```
pipo02mix, line 133: IMO this was a bug preventing rest of code to be executed (without reason)

MarcelMue (approved):
Makes sense to me.

Just for coordination: If either this or #463 merged first, we need to resolve some merge conflicts :)

marians (changes requested):
Currently the changelog entry mentions `--organization` as the new flag, while in code it's `--org`.

I would vote for `--organization` in code, as "organization" is the proper terminology we use and I recommend against introducing another medium-short form in flag names.

cmd/template/cluster/flag.go
```diff
@@ -26,7 +26,8 @@ const (
 	flagMasterAZ            = "master-az" // TODO: Remove some time after August 2021
 	flagName                = "name"
 	flagOutput              = "output"
-	flagOwner               = "owner"
+	flagOrganization        = "org"
```
pipo02mix, line 29:
```suggestion
	flagOrganization        = "organization"
```

marians:
@pipo02mix Did you forget to push? I see no changes.

pipo02mix:
@marians hehe you were right 🙈

cmd/template/nodepool/flag.go
```diff
@@ -117,6 +120,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 	// To be removed around December 2021
 	_ = cmd.Flags().MarkDeprecated(flagClusterIDDeprecated, "use --cluster-name instead")
 	_ = cmd.Flags().MarkDeprecated(flagNodepoolNameDeprecated, "use --description instead")
+	_ = cmd.Flags().MarkDeprecated(flagOwner, "please use --org instead.")
```
marians, line 123:
```suggestion
	_ = cmd.Flags().MarkDeprecated(flagOwner, "please use --organization instead.")
```

cmd/template/cluster/flag.go
```diff
@@ -139,8 +141,18 @@ func (f *flag) Validate() error {
 		}
 	}
 
-	if f.Owner == "" {
-		return microerror.Maskf(invalidFlagError, "--%s must not be empty", flagOwner)
+	// TODO: Remove the flag completely some time after December 2021
+	if f.Owner == "" && f.Organization == "" {
+		return microerror.Maskf(invalidFlagError, "--%s or --%s must not be empty", flagOwner, flagOrganization)
```
marians, line 146: Let's use the desired flag name only here

```suggestion
		return microerror.Maskf(invalidFlagError, "--%s must not be empty", flagOrganization)
```

cmd/template/cluster/flag.go
marians: 🤷

marians (approved):

marians:
Thanks for hanging in there!

PR #474, opened 2021-09-16
Update cluster templating error messages for name flag

kuosandys:
The error messages for templating a cluster are updated to use the `name` flag, replacing the deprecated `cluster-id` flag. The term 'digit' was also incorrectly used to refer to a letter, and has been changed accordingly.


axbarsan (approved):
Really cool ✅

PR #481, opened 2021-09-23
Nodepool nodes are labeled with nodepool id on Azure

fiunchinho:
When creating a nodepool we normally want to label its nodes with the nodepool id so that later we can assign workload to it.
With this change Azure nodepools will be labeled using the label

```
giantswarm.io/machine-pool=a1b2c
```

This is [the label that we mention in our docs](https://docs.giantswarm.io/advanced/node-pools/#assigning-workloads).


whites11 (approved):
YES

PR #484, opened 2021-09-24
Label MachinePool and AzureMachinePool with giantswarm.io/machine-pool

fiunchinho:
I see that we are labeling our nodepool CRs with this label on AWS, so for consistency sake, I'm doing the same on Azure.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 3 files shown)


axbarsan (approved):
Cool

marians (approved):

marians:
Will this annotation also be present if the node pool has been created using a different method, e. g. happa? This should not be a kubectl-gs-only thing.

fiunchinho:
If I understand typescript at all, this was already being done on `happa`. For the [machinepools](https://github.com/giantswarm/happa/blob/master/src/components/MAPI/workernodes/utils.ts#L323) and for the [azuremachinepools](https://github.com/giantswarm/happa/blob/master/src/components/MAPI/workernodes/utils.ts#L271).

PR #485, opened 2021-09-24
Nodepool nodes are labeled with nodepool id on AWS

fiunchinho:
This is a follow up of https://github.com/giantswarm/kubectl-gs/pull/481 since the label was missing from nodes on AWS.

[Our docs say](https://docs.giantswarm.io/advanced/node-pools/#assigning-workloads) that we use the label `giantswarm.io/machine-deployment` on AWS but `giantswarm.io/machine-pool` on Azure. This label can be anything that we want, so I'd suggest to use the same on all providers. This PR uses `giantswarm.io/machine-pool` as that's the value we are already using on Azure, and it makes explicit that we are using CAPI `MachinePools` under the hood. What do you think?

### DIFF
--- a/cmd/template/nodepool/provider/aws.go
+++ b/cmd/template/nodepool/provider/aws.go
@@ -1,6 +1,7 @@
 package provider
 
 import (
+	"fmt"
 	"io"
 	"os"
 	"strconv"
@@ -326,6 +327,10 @@ func newKubeadmConfigFromUnstructured(sshSSOPubKey string, sshdConfig string, o 
 				Groups: &groups,
 				Shell:  &shell,
 			},
+		}
+		kubeadmConfig.Spec.JoinConfiguration.NodeRegistration.KubeletExtraArgs = map[string]string{
+			"cloud-provider": "aws",
+			"node-labels":    fmt.Sprintf("giantswarm.io/machine-pool=%s", o.GetName()),
 		}
 	}
 
--- a/cmd/template/nodepool/provider/aws.go
+++ b/cmd/template/nodepool/provider/aws.go
@@ -7,9 +7,9 @@ import (
 	"strconv"
 	"text/template"
 
-	"github.com/giantswarm/apiextensions/v3/pkg/annotation"
 	"github.com/giantswarm/apiextensions/v3/pkg/apis/infrastructure/v1alpha3"
-	"github.com/giantswarm/apiextensions/v3/pkg/label"
+	"github.com/giantswarm/k8smetadata/pkg/annotation"
+	"github.com/giantswarm/k8smetadata/pkg/label"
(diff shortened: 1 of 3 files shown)


AverageMarcus:
Big +1 for consistency!

Do we want to replace the `giantswarm.io/machine-deployment` completely? (e.g. here: https://github.com/giantswarm/kubectl-gs/blob/1e98bbfa29ecfe443432ca0f19eac4db1a402d10/internal/label/label.go#L10)

I think the `.golden` files will need updating too.

cmd/template/nodepool/provider/aws.go
```diff
@@ -327,6 +328,10 @@ func newKubeadmConfigFromUnstructured(sshSSOPubKey string, sshdConfig string, o
 				Shell:  &shell,
 			},
 		}
+		kubeadmConfig.Spec.JoinConfiguration.NodeRegistration.KubeletExtraArgs = map[string]string{
+			"cloud-provider": "aws",
+			"node-labels":    fmt.Sprintf("giantswarm.io/machine-pool=%s", o.GetName()),
```
njuettner, line 333: Do we want to add this label into apiextensions, it might be we need to pick it up elsewhere?

fiunchinho: good call! Actually, it's already in there.

fiunchinho:
> Big +1 for consistency!
> 
> Do we want to replace the `giantswarm.io/machine-deployment` completely? (e.g. here:
> 
> https://github.com/giantswarm/kubectl-gs/blob/1e98bbfa29ecfe443432ca0f19eac4db1a402d10/internal/label/label.go#L10
> 
> )
> I think the `.golden` files will need updating too.

That'd be a breaking change in our UX. I'd say let's change it when customers migrate to CAPI clusters.

njuettner (approved):
lgtm

anvddriesch (approved):

axbarsan (approved):

marians:
@fiunchinho Thanks! Some questions:

- Are you also going to make the according docs changes?
- Is this working immediately for all releases?
- Will the old annotation `giantswarm.io/machine-deployment` still be set? Could be that customers have automation expecting it.

fiunchinho:
> @fiunchinho Thanks! Some questions:
> 
> * Are you also going to make the according docs changes?
> * Is this working immediately for all releases?
> * Will the old annotation `giantswarm.io/machine-deployment` still be set? Could be that customers have automation expecting it.

I believe we don't have docs for CAPI clusters on GiantSwarm just yet. This is not changing how regular GiantSwarm clusters are templated with `kubectl-gs`, only `CAPA` clusters. So nothing will change for customers, as they are not using CAPA clusters yet.

PR #489, opened 2021-10-01
add get command for releases

AverageMarcus:

- Our public [kubectl-gs documentation](https://docs.giantswarm.io/ui-api/kubectl-gs/) may need an update to reflect the change you are making. - https://github.com/giantswarm/docs/pull/1136

### DIFF
--- a/cmd/get/releases/command.go
+++ b/cmd/get/releases/command.go
@@ -1,0 +1,85 @@
+package releases
+
+import (
+	"io"
+	"os"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/giantswarm/kubectl-gs/pkg/middleware"
+	"github.com/giantswarm/kubectl-gs/pkg/middleware/renewtoken"
+)
+
+const (
+	name  = "releases [release-name]"
+	alias = "release"
+
+	shortDescription = "Display one or more releases"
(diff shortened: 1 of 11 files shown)


cmd/get/releases/command.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/giantswarm/kubectl-gs/pkg/middleware"
+	"github.com/giantswarm/kubectl-gs/pkg/middleware/renewtoken"
+)
+
+const (
+	name  = "releases [release-name]"
```
axbarsan, line 18: Why is this called a `name` and not a `version`?

cmd/get/releases/command.go
```diff
@@ -0,0 +1,85 @@
...
+	alias = "release"
+
+	shortDescription = "Display one or more releases"
+	longDescription  = `Display one or more releases`
+
+	examples = `  # List all releases
+  kubectl gs get releases
+
+  # Get one specific release by its name
+  kubectl gs get release 1.4.2`
```
axbarsan, line 28: This actually requires the `v` prefix to work. Or can we actually make it work both with and without the `v` prefix?
```suggestion
  kubectl gs get release v1.4.2`
```

AverageMarcus: Fixed

cmd/get/releases/command.go
AverageMarcus: No real reason, updated to call it version instead.

axbarsan (approved):

PR #496, opened 2021-10-05
Add templating for EKS clusters using the management cluster API

anvddriesch:
Towards https://github.com/orgs/giantswarm/projects/91

- Our public [kubectl-gs documentation](https://docs.giantswarm.io/ui-api/kubectl-gs/) may need an update to reflect the change you are making.


cmd/template/cluster/provider/templates/aws/template.go
```diff
@@ -7,6 +7,9 @@ import (
 //go:embed cluster.yaml.tmpl
 var cluster string
 
+//go:embed cluster_eks.yaml.tmpl
+var cluster_eks string
```
AverageMarcus, line 11: Is there a reason this uses `_` instead of camel case?

AverageMarcus (approved):
Other than the query about the var naming it LGTM! 👍

cmd/template/cluster/provider/templates/aws/template.go
anvddriesch: Just that `ClusterEKS` is not the object kind. It's just a `Cluster` but the template is different.
I'm fine with using camel case, but I thought it could make it look like that.

AverageMarcus: I've no strong opinions either way though I think as all the other vars here use camel case having this different suggests it has some sort of different purpose or meaning maybe?

anvddriesch: let's go with the camel then :small_airplane:

MarcelMue (approved):

PR #497, opened 2021-10-05
Add support for creating WC keypairs

axbarsan:
Towards https://github.com/giantswarm/roadmap/issues/460


cmd/login/keypair.go
```diff
@@ -0,0 +1,183 @@
...
+			Cert: corev1alpha1.CertConfigSpecCert{
+				AllowBareDomains:    true,
+				ClusterComponent:    keypairUID,
+				ClusterID:           cluster,
+				CommonName:          commonName,
+				DisableRegeneration: true,
+				Organizations:       groups,
+				TTL:                 ttl,
+			},
+			VersionBundle: corev1alpha1.CertConfigSpecVersionBundle{
```
MarcelMue, line 64: I think this entire section is unused in the CRD and can be left out here.

cmd/login/keypair.go
```diff
@@ -0,0 +1,183 @@
...
+	keypairUID := generateKeypairUID()
+	keypairName := fmt.Sprintf("%s-%s", cluster, keypairUID)
+	commonName := fmt.Sprintf("%s.%s.k8s.%s", keypairUID, cluster, clusterBasePath)
+
+	certConfig := &corev1alpha1.CertConfig{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      keypairName,
+			Namespace: metav1.NamespaceDefault,
+			Labels: map[string]string{
+				"cert-operator.giantswarm.io/version": "0.1.0",
```
MarcelMue, line 48: Can this always stay `0.1.0`? I think we have different versions of `cert-operator` or am I mistaken?

cmd/login/keypair.go
```diff
@@ -0,0 +1,183 @@
...
+	return r, nil
+}
+
+// tryToCreateKeypairCredential tries to fetch the keypair credential
+// for a couple of times, until the resource is fetched, or until the timeout is reached.
+func tryToGetKeypairCredential(ctx context.Context, keypairService keypair.Interface, name string) (*corev1.Secret, error) {
+	var secret *corev1.Secret
+	var err error
+
+	retryTimer := time.NewTicker(credentialRetryTimeout)
```
MarcelMue, line 83: I think using the exponential backoff timer we use in operators would simplify this code by a bit.

See this code snippet for example:
https://github.com/giantswarm/aws-operator/blob/49022a4201cda8c6ddf2ebd4bde4c111c4877424/service/internal/ebs/ebs.go#L53-L74

MarcelMue (commented):
Just a quick review in the morning.

marians:
A bunch of feedback regarding user experience. Trying to untangle this as much as possible.

## Mental model

Users should in my opinion understand primarily that in order to execute the command in the right way, they should know whether they want to authenticate with a **management cluster** or a **workload cluster**. So I would make this the most obvious differentiation in the Usage explanation.

As a consequence, the differentiation of the authentication method (OIDC, client certificate, anything else) is secondary / less important.

I want us to design the flags in a way that the order as described above is visible immediately and all flags appear to make sense. Happy to work on the details with you or for you.

## Terminology

Upstream is using the term "client certificate" in documentation ([example](https://kubernetes.io/docs/concepts/security/controlling-access/#authentication)). I didn't find a use of the term "key pair". Since we want to be aligned with upstream when it comes to terminology, I suggest we adopt the term "client certificate" and avoid "key pair".

## Have OIDC for workload clusters in mind

We might soon extend the command to support workload cluster authentication via OIDC. The syntax should ideally be as close as possible.

Again, happy to work on the details.

## Specifying the organization owning the cluster

For ease of use it would be great if `kubectl gs` would find the owner organization name for the user, so that specifying it via a flag would not be necessary. Possibly a future improvement.

If the owner org has to be specified via a flag, the flag name `--keypair-organization` can be a bit confusing to users who are used to [`gsctl create kubeconfig`](https://docs.giantswarm.io/ui-api/gsctl/create-kubeconfig/), as it looks quite similar to the flag `--certificate-organizations`, but means something different.

## Generated context, cluster, and user names

The cluster entry added is currently named `giantswarm-<workload-cluster-name>`. I'd suggest to make it `gs-<management-cluster-codename>-<workload-cluster-name>` (same prefix as MC).

Similarly, the user entry should have a name prefix `gs-`. This will make it easier for users to understand who/what created the entry when looking at their kubeconfig.

Context names are currently named after workload cluster ID and user. For the management cluster we use `gs-<management-cluster-codename>`. I would prefer something similar for workload clusters, as the current naming just isn't very handy for re-using a context. Imagine this:

```
A new kubectl context has been created named 'gs-yy4s5-6f844a96b3fd9df6' and selected.
```

As a user I wouldn't be likely to be able to re-use that context in kubectl once I switched away from it. This instead could work:

```
A new kubectl context has been created named 'gs-ghost-yy4s5' and selected.
```

Also currently we are adding one user and one context entry with every call (of the workload cluster login command). I think that this will become a burden for users. Instead, with both user and context names becoming pattern-based, they should get updated on subsequent calls instead of added/appended.

A user name pattern might be `gs-<mc-name>-<wc-name>-user`. This would allow for only one user per workload cluster by default. We can offer a command option (flag) to set a specific user name to avoid re-using/overwriting a user.

axbarsan:
> ## Mental model
> Users should in my opinion understand primarily that in order to execute the command in the right way, they should know whether they want to authenticate with a **management cluster** or a **workload cluster**. So I would make this the most obvious differentiation in the Usage explanation.
> 
> As a consequence, the differentiation of the authentication method (OIDC, client certificate, anything else) is secondary / less important.
> 
> I want us to design the flags in a way that the order as described above is visible immediately and all flags appear to make sense. Happy to work on the details with you or for you.
> 
> ## Terminology
> Upstream is using the term "client certificate" in documentation ([example](https://kubernetes.io/docs/concepts/security/controlling-access/#authentication)). I didn't find a use of the term "key pair". Since we want to be aligned with upstream when it comes to terminology, I suggest we adopt the term "client certificate" and avoid "key pair".

Can we then have a UX spec taking all these into consideration?

> ## Specifying the organization owning the cluster
> For ease of use it would be great if `kubectl gs` would find the owner organization name for the user, so that specifying it via a flag would not be necessary. Possibly a future improvement.

There can be clusters with the same name across multiple organizations, therefore this is not very straightforward to implement.

> ## Generated context, cluster, and user names
> The cluster entry added is currently named `giantswarm-<workload-cluster-name>`. I'd suggest to make it `gs-<management-cluster-codename>-<workload-cluster-name>` (same prefix as MC).
> 
> Similarly, the user entry should have a name prefix `gs-`. This will make it easier for users to understand who/what created the entry when looking at their kubeconfig.
> 
> Context names are currently named after workload cluster ID and user. For the management cluster we use `gs-<management-cluster-codename>`. I would prefer something similar for workload clusters, as the current naming just isn't very handy for re-using a context. Imagine this:
> 
> ```
> A new kubectl context has been created named 'gs-yy4s5-6f844a96b3fd9df6' and selected.
> ```
> 
> As a user I wouldn't be likely to be able to re-use that context in kubectl once I switched away from it. This instead could work:
> 
> ```
> A new kubectl context has been created named 'gs-ghost-yy4s5' and selected.
> ```
> 
> Also currently we are adding one user and one context entry with every call (of the workload cluster login command). I think that this will become a burden for users. Instead, with both user and context names becoming pattern-based, they should get updated on subsequent calls instead of added/appended.
> 
> A user name pattern might be `gs-<mc-name>-<wc-name>-user`. This would allow for only one user per workload cluster by default. We can offer a command option (flag) to set a specific user name to avoid re-using/overwriting a user.

Ok with these changes.

marians:
FYI added https://github.com/giantswarm/roadmap/issues/486 for the general "key pair" terminology topic

axbarsan:
> > There can be clusters with the same name across multiple organizations, therefore this is not very straightforward to implement.
> 
> Multiple workload clusters with the same name (identifier) in the same installation? That can't work.

Clusters are namespaced by organization, so the name must only be unique within the organization namespace. There can be resources with the same name across multiple namespaces.

marians:
One more thing: the current implementation in this PR set the `CN` field of the generated certificate and does not allow the `kubectl` user to influence it. Here is a resulting example Subject DN:

`Subject: O=6f844a96b3fd9df6, CN=6f844a96b3fd9df6.yy4s5.k8s.ghost.westeurope.azure.gigantic.io`

In the [Rest API ](https://docs.giantswarm.io/api/#operation/addKeyPair) we allow users to set the CN prefix. This is relevant for RBAC, so the client certificate can be mapped to roles based on (cluster) role binding's user subjects.

I think we should add an option for setting the CN prefix here, too.

marians:
Works well for me!

marians:
Another important requirement just came up. In our guide in https://docs.giantswarm.io/ui-api/management-api/wc-key-pairs/ we create the CertConfig with

    cert-operator.giantswarm.io/version: 0.1.0

while in fact this won't work for most clusters.

The version value must match with the cert-operator version that is referenced in the Release CR of the release used for the workload cluster.

PR #470, opened 2021-09-13
'template organization': add --output flag

marians:
Closes https://github.com/giantswarm/giantswarm/issues/18906

This PR adds the `--output` flag, which is present in the other `template` subcommands for specifying the file output, to the `template organization` subcommand.

PR #514, opened 2021-10-12
Add --output flag to 'template app' command

marians:
Same as https://github.com/giantswarm/kubectl-gs/pull/470 but for `template app`.

Closes https://github.com/giantswarm/roadmap/issues/495


marians:
Sounds right.

I'll close this and instead make an issue for removing the `--output` flag.

PR #537, opened 2021-10-28
Use CAPI templates for all releases which name starts with v20.0.0

fiunchinho:
SIG-Product has decided that the first GS CAPI release will be called `v20.0.0-alpha`. Anticipating that it could be renamed in the future, I relaxed the check so that everything works as long as the release name starts with `v20.0.0`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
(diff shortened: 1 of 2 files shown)


MarcelMue (approved):

anvddriesch (approved):

calvix:
isn't better to use semver and compare? once we would need a different version like 	`20.0.1-alpha` this would no longer work

fiunchinho:
> isn't better to use semver and compare? once we would need a different version like `20.0.1-alpha` this would no longer work

I merged too quickly, that's a good point. Let me fix it in another PR.

PR #539, opened 2021-11-02
Use CAPI templates for all releases from v20.0.0-alpha1 onwards

fiunchinho:
Applying suggestions after merging https://github.com/giantswarm/kubectl-gs/pull/537#issuecomment-957465456

### DIFF
--- a/internal/key/key.go
+++ b/internal/key/key.go
@@ -51,6 +51,7 @@ const (
 const (
 	// FirstOrgNamespaceRelease is the first GS release that creates Clusters in Org Namespaces by default
 	FirstAWSOrgNamespaceRelease = "16.0.0"
+	FirstCAPIRelease            = "20.0.0-alpha1"
 )
 
 func BastionResourceName(clusterName string) string {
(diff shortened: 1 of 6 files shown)


internal/key/key.go
```diff
@@ -122,15 +123,11 @@ func GetNodeInstanceProfile(machinePoolID string, clusterID string) string {
...
-	return strings.HasPrefix(version, "20.0.0")
-}
-
-// IsCAPZVersion returns whether a given GS Release Version is based on the CAPI/CAPZ projects
-func IsCAPZVersion(version string) bool {
-	return strings.HasPrefix(version, "20.0.0")
+// IsCAPIVersion returns whether a given GS Release Version uses the CAPI projects
+func IsCAPIVersion(version string) bool {
+	capiVersion, _ := semver.New(FirstCAPIRelease)
+	releaseVersion, _ := semver.New(version)
```
calvix, line 129: stupid question but do we somehow check if the release has the proper format? could there be a scenario where someone inters the wrong semver and this would cause an error here?

calvix: maybe `MustParse` could be better here?

calvix (approved):
lgtm, just minor comment

internal/key/key.go
fiunchinho: I added error handling. Now when using wrong formatted release, this error shows up
```
$ ./kubectl-gs-v1.45.0-dd192b5f87f39e5c13d71e24e204e66ca332dd72-linux-amd64 template cluster --context gs-godsmack --provider azure --release asd --organization giantswarm --description 'e2e test' --output cluster.yaml

Error: Parsing release: No Major.Minor.Patch elements found
```

calvix (approved):
lgtm

PR #540, opened 2021-11-02
Add OpenStack provider

tfussell:
For https://github.com/giantswarm/giantswarm/issues/19400

Adds support for templating OpenStack clusters.

Adds the following OpenStack-specific flags:
- `--cloud`
- `--cloud-config`
- `--dns-nameservers`
- `--external-network-id`
- `--failure-domain`
- `--image-name`
- `--node-machine-flavor`
- `--node-cidr`
- `--root-volume-disk-size`
- `--root-volume-source-type`
- `--root-volume-source-uuid`

Some of these flags will be dropped in favor of defaulting. Others will become flags common for all providers. For now we need the configurability for testing OpenStack alpha releases.


cmd/template/cluster/provider/templates/openstack/open_stack_machine_template.yaml.tmpl
```diff
@@ -0,0 +1,20 @@
...
+    "cluster.x-k8s.io/cluster-name": "{{ .Name }}"
+    "giantswarm.io/organization": "{{ .Organization }}"
+  name: {{ .Name }}
+  namespace: {{ .Namespace }}
+spec:
+  template:
+    spec:
+      flavor: changeme
+      image: changeme
+      sshKeyName: changeme
```
ericgraf, line 16: Pass in as a variable. OpenStack project specific. This is the key stored in OpenStack itself.

cmd/template/cluster/provider/templates/openstack/open_stack_machine_template.yaml.tmpl
```diff
@@ -0,0 +1,20 @@
...
+      cloudName: changeme
```
ericgraf, line 17: Very likely same variable as above cr.  Please pass in .
Default likely will be `openstack`.

cmd/template/cluster/provider/templates/openstack/open_stack_machine_template.yaml.tmpl
```diff
@@ -0,0 +1,20 @@
...
+      identityRef:
+        name: changeme
```
ericgraf, line 19: Very likely same variable as above cr. Please pass in .
Default likely will be `cloud-config`.

cmd/template/cluster/provider/templates/openstack/kubeadm_control_plane.yaml.tmpl
```diff
@@ -0,0 +1,50 @@
...
+          cloud-provider: external
+        name: '{{ `{{ ds.meta_data.hostname }}` }}'
+    joinConfiguration:
+      nodeRegistration:
+        criSocket: /var/run/containerd/containerd.sock
+        kubeletExtraArgs:
+          cloud-provider: external
+        name: '{{ `{{ ds.meta_data.hostname }}` }}'
+    preKubeadmCommands:
+    - hostname "{{ `{{ ds.meta_data.hostname }}` }}"
```
ericgraf, line 33: Why do we have to update the hosts file here ?

cmd/template/cluster/provider/templates/openstack/kubeadm_control_plane.yaml.tmpl
```diff
@@ -0,0 +1,50 @@
...
+    - echo "::1         ipv6-localhost ipv6-loopback" >/etc/hosts
+    - echo "127.0.0.1   localhost" >>/etc/hosts
+    - echo "127.0.0.1   {{ `{{ ds.meta_data.hostname }}` }}" >>/etc/hosts
+    - echo "{{ `{{ ds.meta_data.hostname }}` }}" >/etc/hostname
```
ericgraf, line 37: Cloudinit populates the hostname of the machine. Is there a reason why we had to set this ?

ericgraf (changes requested):

cmd/template/cluster/provider/capo.go
```diff
@@ -0,0 +1,60 @@
...
+
+func WriteCAPOTemplate(ctx context.Context, client k8sclient.Interface, out io.Writer, config ClusterCRsConfig) error {
+	var err error
+
+	homeDir, err := os.UserHomeDir()
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	idRsaPubBytes, err := os.ReadFile(filepath.Join(homeDir, ".ssh/id_rsa.pub"))
```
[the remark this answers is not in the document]
tfussell: Will drop.

cmd/template/cluster/provider/templates/openstack/open_stack_machine_template.yaml.tmpl
```diff
@@ -0,0 +1,24 @@
...
+    "release.giantswarm.io/version": "{{ .ReleaseVersion }}"
+    "giantswarm.io/cluster": "{{ .Name }}"
+    "cluster.x-k8s.io/cluster-name": "{{ .Name }}"
+    "giantswarm.io/organization": "{{ .Organization }}"
+  name: {{ .Name }}
+  namespace: {{ .Namespace }}
+spec:
+  template:
+    spec:
+      flavor: {{ .NodeMachineFlavor }}
```
yulianedyalkova, line 14: Is it `.NodeMachineFlavor` or `.ControlPlaneMachineFlavor` that needs to be set here?

ericgraf: This template is shared for both machine_deployment.yaml.tmpl and kubeadm_control_plane.yaml.tmpl . Does it make sense to have a control plane machine template that is separate to workers?

There are a number of properties that I could see being different down the road.
https://github.com/kubernetes-sigs/cluster-api-provider-openstack/blob/main/config/crd/bases/infrastructure.cluster.x-k8s.io_openstackmachinetemplates.yaml#L73

tfussell: Good point. I want to keep this as simple as possible for now so I will remove `.ControlPlaneMachineFlavor` for now. When we need it we can add a second machine template for the control plane.

cmd/template/cluster/provider/templates/openstack/open_stack_cluster.yaml.tmpl
```diff
@@ -0,0 +1,21 @@
...
+  name: {{ .Name }}
+  namespace: {{ .Namespace }}
+spec:
+  cloudName: changeme
+  identityRef:
+    name: changeme
+    kind: Secret
+  managedAPIServerLoadBalancer: true
+  managedSecurityGroups: true
+  nodeCidr: 10.6.0.0/24
```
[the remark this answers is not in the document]
tfussell: Done.

cmd/template/cluster/flag.go
```diff
@@ -198,6 +246,12 @@ func (f *flag) Validate() error {
 			if len(f.ControlPlaneAZ) > 1 {
 				return microerror.Maskf(invalidFlagError, "--%s supports one availability zone only", flagControlPlaneAZ)
 			}
+		case key.ProviderOpenStack:
+			if f.OpenStack.NodeCIDR != "" {
```
tfussell, line 250: @giantswarm/team-rocket Should we add more validation here?

cmd/template/cluster/provider/templates/aws/cluster.yaml.tmpl
```diff
@@ -12,6 +12,12 @@ metadata:
   name: {{ .Name }}
   namespace: {{ .Namespace }}
 spec:
+{{- if .PodsCIDR }}
```
tfussell, line 15: Made this consistent across providers.

cmd/template/cluster/runner.go
```diff
@@ -88,7 +97,7 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 			return microerror.Mask(err)
 		}
 
-		if r.flag.Provider == key.ProviderAzure {
+		if r.flag.Provider != key.ProviderAWS {
```
tfussell, line 100: Only legacy AWS clusters use the `default` namespace now.

ericgraf (approved):
LGTM!

fiunchinho (approved):

marians:
`kubectl gs template cluster --help` is already pretty messy, because of a lot of options, and I am worried that adding another 12 options will render it useless. That's a big side effect for all users, while the flags to be added will only serve a few.

What could we do?

Can me make the OpenStack related flags hidden, as they are meant to be used by a limited audience only (and may be removed in the future anyway)? This way we wouldn't have negative side effects for people using AWS or Azure only. But that would make the flags harder to use for those few who actually need them.

Can we structure the help output in a way that OpenStack-related flags form their own group? That would actually be a big improvement for every user.

tfussell:
@marians I don't really like the idea of hiding them because I want OpenStack customers to be supported just like AWS and Azure customers.

One option would be to add a subcommand under `opsctl template cluster` for each provider. Then, `opsctl template cluster openstack` could only show the flags for that provider.

Another option would be to parse the `--provider` flag immediately then initialize the other flags depending on its value. `opsctl template cluster --provider openstack --help` would show all flags including OpenStack only while `opsctl template cluster --help` would only show common flags. This isn't very intuitive though.

We could also write our own help message instead of using the automatically generated message.

One last option is to keep it like this until `cobra` merges the PR adding flag grouping support. It sounds like it will be soon. I actually saw your comment there :)

What do you think @giantswarm/sig-ux?

marians:
For reference, here is the upstream issue: https://github.com/spf13/cobra/issues/1327

I can't find a PR to close that issue. The linked PR 1003 implements grouping of commands, not flags.

I would suggest to override the flags help rendering then with our own "hard-coded" string.

cmd/template/cluster/provider/common.go
```diff
@@ -27,7 +27,20 @@ type ClusterCRsConfig struct {
 	EKS                bool
 	ExternalSNAT       bool
 	ControlPlaneSubnet string
-	PodsCIDR           string
+
+	// OpenStack only.
```
MarcelMue, line 31: Are all of these options really unique per workload cluster? 
Wouldn't it make way more sense to default a bunch of these (e.g. SSHKeyName, ImageName, ...) in the webhook as they are tied to the management cluster?

Additionally we know that customers want to use gitops to manage these clusters, supplying them with good overlays seems more sensible to me.

cmd/template/cluster/provider/templates/openstack/machine_deployment.yaml.tmpl
```diff
@@ -0,0 +1,40 @@
+apiVersion: cluster.x-k8s.io/v1alpha4
+kind: MachineDeployment
```
MarcelMue, line 2: Can we use machinepools on Openstack? Machinedeployments are not nice to manage with gitops.

cmd/template/cluster/flag.go
```diff
@@ -64,10 +97,23 @@ func (f *flag) Init(cmd *cobra.Command) {
...
+	cmd.Flags().StringVar(&f.OpenStack.ExternalNetworkID, flagOpenStackExternalNetworkID, "", "External network ID (OpenStack only).")
+	cmd.Flags().StringVar(&f.OpenStack.FailureDomain, flagOpenStackFailureDomain, "", "Failure domain (OpenStack only).")
+	cmd.Flags().StringVar(&f.OpenStack.ImageName, flagOpenStackImageName, "", "Image name (OpenStack only).")
+	cmd.Flags().StringVar(&f.OpenStack.NodeMachineFlavor, flagOpenStackNodeMachineFlavor, "", "Node machine flavor (OpenStack only).")
+	cmd.Flags().StringVar(&f.OpenStack.RootVolumeDiskSize, flagOpenStackRootVolumeDiskSize, "", "Root volume disk size (OpenStack only).")
+	cmd.Flags().StringVar(&f.OpenStack.RootVolumeSourceType, flagOpenStackRootVolumeSourceType, "", "Root volume source type (OpenStack only).")
+	cmd.Flags().StringVar(&f.OpenStack.RootVolumeSourceUUID, flagOpenStackRootVolumeSourceUUID, "", "Root volume source UUID (OpenStack only).")
+	cmd.Flags().StringVar(&f.OpenStack.SSHKeyName, flagOpenStackSSHKeyName, "", "SSH key name (OpenStack only).")
+	cmd.Flags().StringVar(&f.OpenStack.NodeCIDR, flagOpenStackNodeCIDR, "", "CIDR used for the nodes.")
```
marians, line 116: Are any of these flags required with `--provider openstack`? If yes, it would be great if we could mention that here (e. g. instead of `(OpenStack only)` add `Required with provider `openstack`.

marians:
Regarding the general UX concerns, we just discussed that in SIG UX. That should not block this PR in any way. I created this issue to tackle the necessary improvements independently:

- https://github.com/giantswarm/roadmap/issues/605

tfussell:
> I would suggest to override the flags help rendering then with our own "hard-coded" string.

After a team discussion, we decided to simply hide all of the provider-specific flags until we have a better solution.

cmd/template/cluster/provider/common.go
tfussell: I dropped SSH key name because we don't want to allow it. Image name will be decided by us (we're managing images), but is dependent on Kubernetes version and may be useful to override. Source UUID is tricky because it depends on the OpenStack project the cluster is deployed into and that is therefore cluster-dependent.

For the alpha (or "pre-alpha") we just want the ability to specify all of the options. The flags won't show up in the automatic CLI help so I don't think it will be a problem in terms of backwards compatibility. Let me know if you disagree @MarcelMue.

cmd/template/cluster/provider/templates/openstack/machine_deployment.yaml.tmpl
tfussell: We discussed this in Slack.

MarcelMue:
^ sorry for the wrong commit - got mixed up with my CLI.

PR #549, opened 2021-11-08
Make the login `--organization` flag optional

axbarsan:
Closes https://github.com/giantswarm/roadmap/issues/541

This PR makes the `--organization` flag optional when using the `login` command with a workload cluster.
If the `--organization` flag is provided, everything works like before, but if it isn't provided, then:
* we fetch all organizations
* we get the namespaces from their status fields
* we look for a cluster with the given name (`--workload-cluster` flag) in each organization namespace
* if we don't have access to one of the namespaces, we just skip it
* if we only find one, we're good to go
* if we find multiple or none, we throw errors

## Preview

<details>
<summary>Cluster not found</summary>

image

</details>

<details>
<summary>Not enough permissions to get the cluster in the given organization</summary>

image

</details>

<details>
<summary>Multiple clusters with the same name</summary>

image

</details>

### DIFF
--- a/cmd/login/clientcert.go
+++ b/cmd/login/clientcert.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"crypto/sha256"
 	"fmt"
+	"strings"
+	"sync"
 	"time"
 
 	corev1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/core/v1alpha1"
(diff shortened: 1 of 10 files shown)


cmd/login/clientcert.go
```diff
@@ -252,13 +272,89 @@ func fetchCluster(ctx context.Context, clusterService cluster.Interface, provide
...
+
+		return c, nil
+	}
+
+	clustersCh := make(chan *cluster.Cluster, len(namespaces))
+	errorsCh := make(chan error, len(namespaces))
+
+	var wg sync.WaitGroup
+
+	for _, namespace := range namespaces {
```
axbarsan, line 299: A bit of extra spaghetti to execute all cluster fetching requests in parallel

anvddriesch (commented):
If this works I am all good with it. :D
Just curious what made you take this approach. Are there advantages of listing the org namespaces first and then fetching for each of them?
Why not just get one namespace for the desired org (if given) and then list for the namespace? And in case org is not given, use `All` as namespace to list for?

marians:
@anvddriesch My understanding is that non-admin users don't have permission to `list` namespaces (for good reasons). So a query over `all` namespaces fails.

axbarsan:
> Why not just get one namespace for the desired org (if given) and then list for the namespace?

We already do that here:
https://github.com/giantswarm/kubectl-gs/blob/870b996333ae75d6f05a4ad5a8765f254894cfa9/cmd/login/clientcert.go#L285

If the organization name is provided, we only look for the cluster in that organization's namespace.

> And in case org is not given, use `All` as namespace to list for?

In order for that to work, users must have access to listing WCs at the cluster scope, which only works for admin users. Non-admin users have access to listing WCs in one (or more) organization namespaces. Using the new logic we look for the WC in all the org namespaces that the user has access to.

anvddriesch:
Ahhh that makes sense! Thank you for explaining. I assumed that you will simply receive the items in the namespaces that you do have access to.

anvddriesch (approved):

PR #557, opened 2021-11-11
Let users use SA tokens for creating WC client certificates

axbarsan:
Towards https://github.com/giantswarm/roadmap/issues/535

With the changes in this PR, users can now use `ServiceAccount` tokens for users in the `kubectl` config file, and create workload cluster client certificates.

### DIFF
--- a/pkg/kubeconfig/auth.go
+++ b/pkg/kubeconfig/auth.go
@@ -4,6 +4,45 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
 
+type AuthType int
+
+const (
+	AuthTypeUnknown AuthType = iota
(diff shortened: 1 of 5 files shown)


pkg/kubeconfig/auth.go
```diff
@@ -4,6 +4,45 @@ import (
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
 )
 
+type AuthType int
+
+const (
+	AuthTypeUnknown AuthType = iota
+	AuthTypeServiceAccount
+	AuthTypeAuthProvider
+	AuthTypeClientCertificate
```
axbarsan, line 13: Not every type out there, but we can add them as we go

anvddriesch (approved):
Nice! :smile:

PR #564, opened 2021-11-12
add self contained kubeconfig flag

anvddriesch:
Towards https://github.com/giantswarm/roadmap/issues/542

- Our public [kubectl-gs documentation](https://docs.giantswarm.io/ui-api/kubectl-gs/) may need an update to reflect the change you are making.

### DIFF
--- a/cmd/login/clientcert.go
+++ b/cmd/login/clientcert.go
@@ -15,11 +15,11 @@ import (
 	kubeconfig "github.com/giantswarm/kubeconfig/v3"
 	"github.com/giantswarm/microerror"
 	"github.com/spf13/afero"
+	yaml "gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
-	"sigs.k8s.io/yaml"
 
 	"github.com/giantswarm/kubectl-gs/internal/feature"
 	"github.com/giantswarm/kubectl-gs/internal/key"
@@ -236,7 +236,7 @@ func printCredential(fs afero.Fs, filePath string, clientCert *clientcert.Client
 		},
 		Contexts: []kubeconfig.KubeconfigNamedContext{
 			{
-				Name: fmt.Sprintf("%s-context", contextName),
+				Name: contextName,
 				Context: kubeconfig.KubeconfigContext{
 					Cluster: contextName,
 					User:    fmt.Sprintf("%s-user", contextName),
--- a/cmd/login/clientcert.go
+++ b/cmd/login/clientcert.go
@@ -13,7 +13,6 @@ import (
 	"github.com/giantswarm/backoff"
 	"github.com/giantswarm/microerror"
 	"github.com/spf13/afero"
-	yaml "gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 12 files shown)


marians:
Looks good!

One nitpick I have is the unnecessary empty `certificate-authority` attribute.

image

I guess that's a problem in our kubeconfig library. I think `omitempty` is missing here: https://github.com/giantswarm/kubeconfig/blob/master/types.go#L42-L43

marians:
I just noticed that we are not checking whether an existing output file gets overwritten. Let's please do that early, and show an error if the file already exist.

cmd/login/clientcert.go
```diff
@@ -11,8 +12,10 @@ import (
 	corev1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/core/v1alpha1"
 	"github.com/giantswarm/apiextensions/v3/pkg/label"
 	"github.com/giantswarm/backoff"
+	kubeconfig "github.com/giantswarm/kubeconfig/v3"
```
axbarsan, line 15: I don't think we need this.
`"k8s.io/client-go/tools/clientcmd/api"` provides everything we need and we already use it in multiple places (even in this file)

cmd/login/runner.go
```diff
@@ -415,7 +415,7 @@ func (r *runner) createClusterClientCert(ctx context.Context) error {
 	}
 
 	// Store client certificate credential into the kubeconfig.
-	contextName, contextExists, err := storeCredential(r.k8sConfigAccess, r.fs, clientCertResource, secret, clusterBasePath)
+	contextName, contextExists, err := storeCredential(r.k8sConfigAccess, r.fs, r.flag.WCSelfContained, clientCertResource, secret, clusterBasePath)
```
axbarsan, line 418: Is it possible to have an if statement here where we decide whether we `store` or `print` the credential? It's a bit confusing to have printing done in the `storeCredential` function

anvddriesch: Sure, we can do that, too! :)

cmd/login/clientcert.go
anvddriesch: Cool, I'll use that one instead!

cmd/login/clientcert.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/giantswarm/backoff"
 	"github.com/giantswarm/microerror"
 	"github.com/spf13/afero"
+	yaml "gopkg.in/yaml.v2"
```
axbarsan, line 16: Sorry, I missed this before. Can we use `sigs.k8s.io/yaml`? It's mentioned in [fmt](https://github.com/giantswarm/fmt/blob/master/go/yaml_marshalling.md)

anvddriesch: Yeah I also just noticed that. I copied this function from gsctl so the older dependencies came along.

cmd/login/clientcert.go
```diff
@@ -245,16 +244,12 @@ func printCredential(k8sConfigAccess clientcmd.ConfigAccess, fs afero.Fs, filePa
...
-	if err != nil {
-		return "", false, microerror.Mask(err)
-	}
 	if exists, err := afero.Exists(fs, filePath); exists {
 		return "", false, microerror.Maskf(fileExistsError, "The destination file %s already exists. Please specify a different destination.", filePath)
 	} else if err != nil {
 		return "", false, microerror.Mask(err)
 	}
-	err = afero.WriteFile(fs, filePath, bytes, 0600)
+	err = clientcmd.WriteToFile(kubeconfig, filePath)
```
axbarsan, line 252: Nice 👍

axbarsan (approved):

PR #571, opened 2021-11-17
Add URL login support for ServiceAccount tokens in the `login` command

axbarsan:
This PR adds the ability of executing the MC login part of the `login` command with a `ServiceAccount` token (using the `--token` flag). This is very useful in automation, so you don't have to hard-code MC login information.

This allows doing a WC login in automation using a really short command, like
```
kubectl gs login https://g8s.mymc.gigantic.io \
--token <SA-TOKEN> \
--workload-cluster <CLUSTER-NAME> \
--certificate-group some-group
```
without needing any pre-existing `kubectl` context.

### DIFF
--- a/cmd/login/oidc.go
+++ b/cmd/login/oidc.go
@@ -1,0 +1,159 @@
+package login
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"time"
+
+	gooidc "github.com/coreos/go-oidc/v3/oidc"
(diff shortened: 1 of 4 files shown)


cmd/login/oidc.go
```diff
@@ -0,0 +1,159 @@
+package login
```
axbarsan, line 1: Moved this part into a separate file, since it's now also possible to authenticate with an SA token

anvddriesch (approved):
Very nice :)

PR #581, opened 2021-11-22
Add flag for running the login command with a cluster in the default namespace

axbarsan:
This is only for GS staff, as this has been quite an annoying issue for on-call people: not being able to create client certificates for clusters that have not been migrated to organization namespaces.

This PR allows looking for a cluster in the `default` namespace if using the hidden `--insecure-namespace` flag


anvddriesch (approved):
sweet!

PR #585, opened 2021-11-23
Disable version caching for the `selfupdate` command

axbarsan:
Our self-updating logic caches the latest version for an hour before checking for new versions again. This is useful because we check for updates when any command runs. However, the version shouldn't be cached when running the `selfupdate` command, as users should be able to get the latest version right away.


puja108 (approved):
<3

marians (approved):
Makes sense!

PR #586, opened 2021-11-24
Default release version in the template cluster/nodepool subcommands

axbarsan:
As seen on https://gigantic.slack.com/archives/C02EVLE9W/p1637762623395000

The `--release` flag was accidentally made required by some internal change. This PR makes sure we always have a default value:
- for `template cluster`, we use the latest active release in the MC
- for `template nodepool`, we use the workload cluster's release version

**BONUS:** We now validate if the cluster exists when running `template nodepool`.

### DIFF
--- a/cmd/template/nodepool/runner.go
+++ b/cmd/template/nodepool/runner.go
@@ -7,10 +7,13 @@ import (
 	"strings"
 
 	"github.com/giantswarm/apiextensions/v3/pkg/id"
+	"github.com/giantswarm/apiextensions/v3/pkg/label"
 
 	"github.com/giantswarm/kubectl-gs/cmd/template/nodepool/provider"
 	"github.com/giantswarm/kubectl-gs/internal/key"
 	"github.com/giantswarm/kubectl-gs/pkg/commonconfig"
+	dataClient "github.com/giantswarm/kubectl-gs/pkg/data/client"
+	"github.com/giantswarm/kubectl-gs/pkg/data/domain/cluster"
 
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
@@ -47,6 +50,12 @@ func (r *runner) Run(cmd *cobra.Command, args []string) error {
 func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) error {
 	var err error
 
+	commonConfig := commonconfig.New(r.flag.config)
+	c, err := commonConfig.GetClient(r.logger)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
 	var config provider.NodePoolCRsConfig
 	{
 		config = provider.NodePoolCRsConfig{
@@ -72,9 +81,6 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 			config.NodePoolID = id.Generate()
 		}
 
-		// Remove leading 'v' from release flag input.
-		config.ReleaseVersion = strings.TrimLeft(config.ReleaseVersion, "v")
-
 		if len(r.flag.AvailabilityZones) > 0 {
 			config.AvailabilityZones = r.flag.AvailabilityZones
 		}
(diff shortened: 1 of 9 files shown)


cmd/template/nodepool/runner.go
```diff
@@ -123,3 +133,41 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
...
+	} else if err != nil {
+		return nil, microerror.Mask(err)
+	}
+
+	switch wc := c.(type) {
+	case *cluster.Cluster:
+		return wc, nil
+	case *cluster.Collection:
+		targetWC := wc.Items[0]
+		return &targetWC, nil
```
axbarsan, line 169: Not sure if `&wc.Items[0]` would be safe here, as I believe `wc` would go out of scope and be garbage collected. Or would it be borrowed?

cmd/template/nodepool/runner.go
```diff
@@ -85,12 +91,16 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
...
-	if err != nil {
-		return microerror.Mask(err)
+		wc, err := getWorkloadCluster(ctx, c, config.Namespace, config.ClusterName, r.flag.Provider)
+		if err != nil {
+			return microerror.Mask(err)
+		}
+		if len(config.ReleaseVersion) < 1 {
+			config.ReleaseVersion = wc.Cluster.Labels[label.ReleaseVersion]
+		} else {
+			config.ReleaseVersion = strings.TrimLeft(config.ReleaseVersion, "v")
```
anvddriesch, line 102: I'm not even sure we need an `else` case here. I can not think of any situation where we would want the nodepool to be a different version than the cluster.

cmd/template/nodepool/runner.go
njuettner: maybe adding a test for this function would make sense, wdyt?

whites11:
> **BONUS:** We now validate if the cluster exists when running `template nodepool`.

this means you cannot create the template for both a cluster and a node pool and create them with one single `kubectl apply` call?

cmd/template/nodepool/runner.go
axbarsan: The `else` branch is used when the user provides the `--release` flag, and the value isn't defaulted.

If we want to remove the flag, it should probably be deprecated first for some time

axbarsan:
> > **BONUS:** We now validate if the cluster exists when running `template nodepool`.
> 
> this means you cannot create the template for both a cluster and a node pool and create them with one single `kubectl apply` call?

Hmmm, good point. When there are multiple resource templates in a YAML, aren't they created in order, rather than all of them at the same time?

whites11:
> > > **BONUS:** We now validate if the cluster exists when running `template nodepool`.
> > 
> > 
> > this means you cannot create the template for both a cluster and a node pool and create them with one single `kubectl apply` call?
> 
> Hmmm, good point. When there are multiple resource templates in a YAML, aren't they created in order, rather than all of them at the same time?

Yes, but If I run:

```
kubectl template cluster ...
kubectl template nodepool .. <fails because cluster is not applied
```

axbarsan:
> > > > **BONUS:** We now validate if the cluster exists when running `template nodepool`.
> > > 
> > > 
> > > this means you cannot create the template for both a cluster and a node pool and create them with one single `kubectl apply` call?
> > 
> > 
> > Hmmm, good point. When there are multiple resource templates in a YAML, aren't they created in order, rather than all of them at the same time?
> 
> Yes, but If I run:
> 
> ```
> kubectl template cluster ...
> kubectl template nodepool .. <fails because cluster is not applied
> ```

Ah right. So we can change the implementation so the cluster would only be fetched if the release version is not provided.
My only concern with this is I'm not sure if it won't be confusing to only have the command fail if the `--release` flag is not provided and the WC does not exist.

cmd/template/nodepool/runner.go
axbarsan: Added one now, please take a look

fiunchinho (changes requested):
I know I proposed checking the cluster's release when creating a node pool but after thinking about it, I'm not sure it's the right approach anymore. 
By doing it this way, we prevent customers from generating their templates before applying them. Customers wouldn't be able to generate all their templates for both clusters and nodepools without having to apply the cluster template first.

I think this will make it harder for us to follow a gitops approach where all the manifests would be pushed to a git repository to be applied later on. I believe the release flag should be mandatory, or defaulted to the latest active release.

What do you think?

axbarsan:
> I think this will make it harder for us to follow a gitops approach where all the manifests would be pushed to a git repository to be applied later on. I believe the release flag should be mandatory, or defaulted to the latest active release.

For clusters we already default to the latest active release. For node pools that wouldn't be possible, as there could be a mismatch between the node pool release and the cluster release.

We can make the flag mandatory, although it is a bit weird to have it like that for `template nodepool`, where it has no effect (execept for templating), but still have it optional for `template cluster`, where it actually has effect.

The way to go would probably be making the flag mandatory in both commands

axbarsan:
> The fact the release is matching is checked by admission controllers. I would make it optional for both clusters and node pools and if the customer specifies two different versions it's their problem

We cannot leave it optional, as the commands determine based on it whether if it should template CAPI or GS resources. The user could template GS resources and the admission controller could default to a CAPI release.

fiunchinho:
Then let's make it mandatory for both commands. What do you think @marians ?

axbarsan:
Closed in favour of https://github.com/giantswarm/kubectl-gs/pull/589

PR #595, opened 2021-12-01
When generating a kubeconfig, look for the generated secret in all namespaces

whites11:
Currently there is an hardcoded logic in `kubectl gs login` that looks for the secret generated by cert-operator in the `default` namespace for azure clusters.
With the bump to latest cert-operator on the newest azure release, this won't work any more (the secret is generated in the org namespace).

Problem: we're going to have a mixed situation for azure, until all clusters are migrated to the newest release (yet to be released).

With this PR I suggest relaxing the namespace check and looking for the desired secret by label.

Blocks https://github.com/giantswarm/roadmap/issues/554

### DIFF
--- a/pkg/data/domain/clientcert/service.go
+++ b/pkg/data/domain/clientcert/service.go
@@ -61,10 +61,17 @@ func (s *Service) Delete(ctx context.Context, clientCert *ClientCert) error {
 }
 
 func (s *Service) GetCredential(ctx context.Context, namespace, name string) (*corev1.Secret, error) {
-	secret, err := s.client.K8sClient.K8sClient().CoreV1().Secrets(namespace).Get(ctx, name, metav1.GetOptions{})
+	secret := &corev1.Secret{}
+	err := s.client.K8sClient.CtrlClient().Get(ctx, runtimeclient.ObjectKey{Name: name, Namespace: namespace}, secret)
 	if apierrors.IsNotFound(err) {
-		return nil, microerror.Mask(notFoundError)
-	} else if err != nil {
+		// Try in default namespace for legacy azure clusters.
+		err = s.client.K8sClient.CtrlClient().Get(ctx, runtimeclient.ObjectKey{Name: name, Namespace: metav1.NamespaceDefault}, secret)
+		if apierrors.IsNotFound(err) {
+			return nil, microerror.Mask(notFoundError)
+		}
+	}
+
+	if err != nil {
 		return nil, microerror.Mask(err)
 	}
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 - Fix CRD API version for EKS infrastcture ref in cluster CR template. 
 
+### Changed
+
+- When generating a kubeconfig, look for the generated secret in org namespace and fallback to default namespace to support legacy Azure clusters.
+
 ## [1.53.0] - 2021-11-29
 
 ### Changed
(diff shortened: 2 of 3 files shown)


pkg/data/domain/clientcert/service.go
```diff
@@ -60,13 +64,26 @@ func (s *Service) Delete(ctx context.Context, clientCert *ClientCert) error {
...
-	if apierrors.IsNotFound(err) {
-		return nil, microerror.Mask(notFoundError)
-	} else if err != nil {
+func (s *Service) GetCredential(ctx context.Context, clusterId, certUid string) (*corev1.Secret, error) {
+	selector := map[string]string{
+		kgslabel.Certificate: certUid,
+		apiextlabel.Cluster:  clusterId,
+	}
+
+	secrets, err := s.client.K8sClient.K8sClient().CoreV1().Secrets("").List(ctx, metav1.ListOptions{
```
fiunchinho, line 73: for consistency with other places in the code, could you use the controller client rather than the typed clientset?

whites11: Are we using it elsewhere in here? I just used what was there before. But yeah let's change it

fiunchinho: On the same file I see we are using it

whites11: Good will fix it thx

anvddriesch (commented):
I agree that we need this. However, what I wonder is... what happens when a user does not have permission to list all namespaces. Will the login command still work for them?

whites11:
> I agree that we need this. However, what I wonder is... what happens when a user does not have permission to list all namespaces. Will the login command still work for them?

Good point. Any idea how to quickly test this scenario?

anvddriesch:
I am actually not sure. Maybe creating another rbac role? @giantswarm/team-rainbow what would be the simplest way of testing something with differenct access?

pkg/data/domain/clientcert/service.go
whites11: done

marians:
My suggestion for testing:

1. Pick a test installation and a test organization
2. Create a service account in the test org's namespace. Bind that service account to the "admin" ClusterRole (in that namespace). You can do this in one step using happa's Access Control UI under Organizations.
3. Get the service account's token.
4. Use `kubectl gs login <endpoint> --workload-cluster ...` with that service account token (you'll have to set kubectl flags like `--token` and probably more).

whites11:
It seems to be broken already. I have a service account as Marian suggested, but I can't create a kubeconfig with latest kubectl-gs:

```
$ kgs login gremlin --workload-cluster a7g2w --certificate-group system:masters --certificate-ttl 8h --token $TOKEN --organization gerald

Context 'gs-gremlin' is already selected.
You are logged in to the management cluster of installation 'gremlin'.
Error: Organizations.security.giantswarm.io "gerald" is forbidden: User "system:serviceaccount:org-gerald:gerald-and-friends" cannot get resource "organizations" in API group "security.giantswarm.io" at the cluster scope
```

whites11:
According to rainbow we can assume customers have cluster-admin permissions so far. So I guess permissions are not a problem. See https://gigantic.slack.com/archives/C02FJACQ3D4/p1638441970101300

fiunchinho (approved):

anvddriesch:
I am okay with supporting this case, but can we limit looking in all namespaces to the affected azure releases?
Since we want to work towards allowing more access control for the customers.

whites11:
OK to be on the safe side, I changed the approach.
Now the secret is looked for as it was before with a Get in the same namespace as the CertConfig.
In case it is not found, we try to look for it in the `default` namespace as well.
This should require the same permissions as before or fail the same way.

anvddriesch (approved):
Nice :)

PR #598, opened 2021-12-02
Support CAPO K8s API URL scheme

kopiczko:
Towards https://github.com/giantswarm/roadmap/issues/584

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
+- Add support for the new URL scheme `INSTALLATION.OWNER_ID.gigantic.io`.
+
 ## [1.53.0] - 2021-11-29
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ### Added
 
-- Add support for the new URL scheme `INSTALLATION.OWNER_ID.gigantic.io`.
+- Add support for the new URL scheme `api.INSTALLATION.OWNER_ID.gigantic.io`.
 
 ## [1.53.0] - 2021-11-29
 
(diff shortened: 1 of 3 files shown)


pkg/installation/url.go
```diff
@@ -35,7 +35,8 @@ func GetUrlType(u string) int {
 }
 
 func isK8sApiUrl(u string) bool {
-	return k8sApiURLRegexp.MatchString(u)
+	u = strings.SplitN(u, ":", 2)[0]
+	return k8sApiURLRegexp.MatchString(u) || (strings.HasPrefix(u, "api") && strings.HasSuffix(u, "gigantic.io"))
```
tfussell, line 39: Probably doesn't matter that much, but this would match invalid strings such as `apigigantic.io` or `api.gigantic.io`.

pkg/installation/url.go
```diff
@@ -35,7 +35,8 @@ func GetUrlType(u string) int {
 }
 
 func isK8sApiUrl(u string) bool {
-	return k8sApiURLRegexp.MatchString(u)
+	u = strings.SplitN(u, ":", 2)[0]
```
tfussell, line 38: This is removing the port? Might be more robust to use `url.Parse`.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
+- Add support for the new URL scheme `api.INSTALLATION.OWNER_ID.gigantic.io`.
```
marians, line 12: Can we make this description talk to end users? It should be clear which commands are affected, and potentially the release and provider.

kopiczko: This affects all future providers and I don't think we'll change URL scheme for a specific release.  You can clearly see if URL for the installation matches the pattern. How this can be clearer for the end user?

pkg/installation/url.go
kopiczko: I feel like this is robust enough

tfussell (approved):

CHANGELOG.md
marians: It must somehow be possible to explain which commands are affected in what way, no?

kopiczko: https://github.com/giantswarm/kubectl-gs/pull/606

PR #613, opened 2021-12-09
Schedule cluster update

njuettner:
We want to allow users scheduling cluster updates via `kubectl-gs`.

I did try to find a place where we can support this. Currently we have `kubectl gs update app` and I added `kubectl gs update cluster` to that. If there's a better proposal I'm totally open about that.

Example:

Scheduling cluster updates
```
kubectl gs update cluster --name abcd1 --namespace my-org --release-version 16.1.0 --scheduled-time "2022-01-30 12:00" --provider aws
```

Update cluster immediately
```
kubectl gs update cluster --name abcd1 --namespace my-org --release-version 16.1.0 --provider aws
```

Issue: https://github.com/giantswarm/giantswarm/issues/19989

### DIFF
--- a/cmd/update/cluster/command.go
+++ b/cmd/update/cluster/command.go
@@ -23,16 +23,19 @@ const (
 Updates given cluster with the provided values.
 
 Options:
-  --name <name>               		Name of the cluster to update.
-  --namespace <namespace>     		Namespace of the cluster.
+  --name <cluster-name>             	Name of the cluster to update.
+  --namespace <cluster-namespace>   	Namespace of the cluster.
   --release-version <release-version>   Update the cluster to a release version. The release version must be higher than the current release version.
-  --scheduled-time <scheduled-time>     Scheduled time when cluster should be updated. The value has to be in RFC822 Format and UTC time zone.
+  --scheduled-time <scheduled-time>     Optionally: Scheduled time when cluster should be updated. The value has to be in RFC822 Format and UTC time zone.
   --provider <provider> 		Name of the provider.`
 
 	examples = `  # Display this help
 kubectl gs update cluster --help
 
-# Update cluster version
+# Update cluster
+kubectl gs update cluster --name abcd1 --namespace my-org --release-version 16.1.0 --provider aws
+
+# Schedule cluster update
 kubectl gs update cluster --name abcd1 --namespace my-org --release-version 16.1.0 --scheduled-time "30 Jan 22 12:00 UTC" --provider aws`
 )
 
--- a/cmd/update/cluster/command.go
+++ b/cmd/update/cluster/command.go
@@ -27,7 +27,7 @@ Options:
   --namespace <namespace>     		Namespace of the cluster.
   --release-version <release-version>   Update the cluster to a release version. The release version must be higher than the current release version.
   --scheduled-time <scheduled-time>     Scheduled time when cluster should be updated. The value has to be in RFC822 Format and UTC time zone.
-  --provider <provider> 			Name of the provider.`
+  --provider <provider> 		Name of the provider.`
 
 	examples = `  # Display this help
 kubectl gs update cluster --help
--- a/cmd/update/cluster/command.go
+++ b/cmd/update/cluster/command.go
@@ -1,0 +1,88 @@
+package cluster
+
+import (
+	"io"
+	"os"
+
+	"github.com/giantswarm/microerror"
+	"github.com/giantswarm/micrologger"
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/giantswarm/kubectl-gs/pkg/middleware"
+	"github.com/giantswarm/kubectl-gs/pkg/middleware/renewtoken"
+)
+
+const (
+	name = "cluster --name <cluster-name> --namespace <cluster-namespace> --release-version <release-version> --scheduled-time <scheduled-time> --provider <provider>"
+
+	shortDescription = "Schedule a cluster update."
+	longDescription  = `Schedule a cluster update.
+
+Updates given cluster with the provided values.
+
+Options:
+  --name <name>               		Name of the cluster to update.
+  --namespace <namespace>     		Namespace of the cluster.
+  --release-version <release-version>   Update the cluster to a release version. The release version must be higher than the current release version.
+  --scheduled-time <scheduled-time>     Scheduled time when cluster should be updated. The value has to be in RFC822 Format and UTC time zone.
+  --provider <provider> 			Name of the provider.`
+
+	examples = `  # Display this help
+kubectl gs update cluster --help
+
+# Update cluster version
+kubectl gs update cluster --name abcd1 --namespace my-org --release-version 16.1.0 --scheduled-time "30 Jan 22 12:00 UTC" --provider aws`
+)
+
+type Config struct {
+	Logger     micrologger.Logger
+	FileSystem afero.Fs
+
+	K8sConfigAccess clientcmd.ConfigAccess
+
+	Stderr io.Writer
+	Stdout io.Writer
+}
+
+func New(config Config) (*cobra.Command, error) {
+	if config.Logger == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.Logger must not be empty", config)
+	}
+	if config.K8sConfigAccess == nil {
+		return nil, microerror.Maskf(invalidConfigError, "%T.K8sConfigAccess must not be empty", config)
+	}
+	if config.Stderr == nil {
+		config.Stderr = os.Stderr
+	}
+	if config.Stdout == nil {
+		config.Stdout = os.Stdout
+	}
+
+	f := &flag{}
+
+	r := &runner{
+		flag:   f,
+		logger: config.Logger,
+
+		stderr: config.Stderr,
+		stdout: config.Stdout,
+	}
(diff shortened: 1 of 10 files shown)


cmd/update/cluster/command.go
```diff
@@ -0,0 +1,88 @@
...
+	longDescription  = `Schedule a cluster update.
+
+Updates given cluster with the provided values.
+
+Options:
+  --name <name>               		Name of the cluster to update.
+  --namespace <namespace>     		Namespace of the cluster.
+  --release-version <release-version>   Update the cluster to a release version. The release version must be higher than the current release version.
+  --scheduled-time <scheduled-time>     Scheduled time when cluster should be updated. The value has to be in RFC822 Format and UTC time zone.
+  --provider <provider> 			Name of the provider.`
```
njuettner, line 30: Unfortunately provider has to be provided, this is a design choice which was made in `pkg/data/domain/cluster`, I'm not happy with that but I don't want to rewrite the whole code base.

When we fetch a cluster via `get` function we automatically fetch the infra provider CR even you don't need it but that's how it is designed.

cmd/update/cluster/runner.go
```diff
@@ -0,0 +1,126 @@
...
+
+	err = r.run(ctx, cmd, args)
+	if err != nil {
+		return microerror.Mask(err)
+	}
+
+	return nil
+}
+
+func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) error {
```
njuettner, line 45: We do not have to check if release version or schedule time might be not valid because we have for both providers (AWS and Azure) validation checks which would block users from using the wrong schedule format or invalid release version.

cmd/update/cluster/command.go
```diff
@@ -0,0 +1,88 @@
...
+	"github.com/spf13/afero"
+	"github.com/spf13/cobra"
+	"k8s.io/client-go/tools/clientcmd"
+
+	"github.com/giantswarm/kubectl-gs/pkg/middleware"
+	"github.com/giantswarm/kubectl-gs/pkg/middleware/renewtoken"
+)
+
+const (
+	name = "cluster --name <cluster-name> --namespace <cluster-namespace> --release-version <release-version> --scheduled-time <scheduled-time> --provider <provider>"
```
whites11, line 18: can we maybe use a better const name here? `name` sounds wrong to me.

whites11 (approved):

cmd/update/cluster/flag.go
```diff
@@ -0,0 +1,61 @@
...
+type flag struct {
+	config         genericclioptions.RESTClientGetter
+	print          *genericclioptions.PrintFlags
+	Name           string
+	ReleaseVersion string
+	ScheduledTime  string
+	Provider       string
+}
+
+func (f *flag) Init(cmd *cobra.Command) {
```
anvddriesch, line 25: I think the `--namespace` flag is missing here.

cmd/update/cluster/flag.go
```diff
@@ -0,0 +1,61 @@
...
+	cmd.Flags().StringVar(&f.Name, flagName, "", "Name of the cluster to update.")
+	_ = cmd.Flags().MarkHidden(flagName)
```
anvddriesch, line 27: Since all these flags are hidden, am I correct to assume that this command is mainly for GS staff to use?
This is a very useful command. 
What do you think about also supporting the "instant" case of just updating the label without `--scheduled-time` like we have in happa? Or do you think we should rather encourage scheduling?

cmd/update/cluster/flag.go
njuettner: This is some magic because we use `kubectl` which I recently learned it's injected magically, so it we don't have to specify it.

cmd/update/cluster/flag.go
njuettner: Make sense, so shall `--scheduled-time` be optional and without specifying we update the release version directly?

cmd/update/cluster/flag.go
anvddriesch: Oh right. :) It slipped my mind.

cmd/update/cluster/flag.go
anvddriesch: If it's not too much effort to include, I think it would be nice. :)

cmd/update/cluster/command.go
njuettner: This is the same const name like any other subcommand description, would it be ok for you to align with the other ones? Otherwise we need to change it everywhere.

Here's what I mean some examples: 
- https://github.com/giantswarm/kubectl-gs/blob/master/cmd/update/app/command.go#L18
- https://github.com/giantswarm/kubectl-gs/blob/master/cmd/login/command.go#L18

cmd/update/cluster/flag.go
njuettner: Ok done 🙂 , PTAL

anvddriesch (approved):
Nice!

cmd/update/cluster/command.go
whites11: all good

marians:
Nice addition in general! Some comments:

Why is there this `--provider` flag? If we can find out programmatically what provider the installation/cluster is using, then the user should not have to define this. I am certain we can.

The date format appears unfortunate to me. If we would ever have to chose a format, we would likely Chose a date format that is less ambiguous. For usability sake, YYYY-MM-DD HH:MM:SS is by far less error-prone. Some tests:

- What's the ISO representation for this date `22 Jan 21 01:00 UTC? ? What is day and what is year?
- Is `1 Jan 2021 01:00 UTC` valid, where the date is single digit and the year uses 4 digits?
- Is the month case sensitive?
- Can I use different time zone identifiers? And if yes, which are valid?

I would strongly vote for accepting these two formats only:

- `YYYY-MM-DD HH:MM`
- `YYYY-MM-DD HH:MM:SS`

And as a time zone we should always assume UTC and not allow anything else.

As a result of the command, we can print the given date in a more verbose format, using the locale of the user's machine. Example:

```
An upgrade of cluster abcd1 to release v16.1.0 has been scheduled for

    Sunday, 30 January 2022, 12:00 UTC (Sunday, 30 January 2022, 13:00 CET)
```

marians:
(Regarding implementation: we can of course convert the input date to the necessary format for the annotation. No need to change that.)

njuettner:
Thanks @marians good feedback 👍🏻.

The reason for the provider flag is because we use different labels (we still have some clusters which are using the older cluster labels on AWS).

On AWS older clusters only have `giantswarm.io/cluster` whereas in newer clusters on AWS and Azure we have `cluster.x-k8s.io/cluster-name`. So we have to fetch the clusters differently.

I checked the clusters on AWS and it seems only release 14.x uses the older labels `giantswarm.io/cluster`. Those releases are also deprecated and I could manually apply the new selector which would be the same for all providers.

WDYT?

The date format sounds good to me and is better I agree. I'll try to change that 👍🏻.
Admission webhook only allows UTC there's no other format supported.

njuettner:
Ok so the time format is changed, I followed your suggestions. Users can schedule clusters via `YYYY-MM-DD HH:MM`. Seconds are ignored by `ISO 8601` which is used in the annotation so I dropped them.

Prompted message for cluster schedule look like this example (`RFC1123`):
```
An upgrade of cluster abcd1 to release 16.1.0 has been scheduled for

		Sat, 01 Jan 2022 01:00:00 UTC, (Sat, 01 Jan 2022 02:00:00 CET)
```

cmd/update/cluster/runner.go
```diff
@@ -0,0 +1,161 @@
...
+					Op:    "add",
+					Path:  fmt.Sprintf("/metadata/annotations/%s", replaceToEscape(annotation.UpdateScheduleTargetTime)),
+					Value: scheduledTime,
+				},
+			},
+		}
+		messageFormat := `An upgrade of cluster %s to release %s has been scheduled for
+
+		%v, (%v)
+		`
```
marians, line 114: Let's use spaces instead of tabs. That's more predictable in terms of the actual rendering.

```suggestion
		messageFormat := "An upgrade of cluster %s to release %s has been scheduled for\n    %v, (%v)"
```

marians:
Thanks for handling my feedback so nicely! ❤️ 

> I checked the clusters on AWS and it seems only release 14.x uses the older labels giantswarm.io/cluster. Those releases are also deprecated and I could manually apply the new selector which would be the same for all providers.

> WDYT?

I don't really know what this means. If you like, we can get into a call.

cmd/update/cluster/runner.go
```diff
@@ -108,10 +108,7 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 				},
 			},
 		}
-		messageFormat := `An upgrade of cluster %s to release %s has been scheduled for
-
-		%v, (%v)
-		`
+		messageFormat := "An upgrade of cluster %s to release %s has been scheduled for\n    %v, (%v)"
```
marians, line 111:
```suggestion
		messageFormat := "An upgrade of cluster %s to release %s has been scheduled for\n\n    %v, (%v)"
```

Sorry, forgot one newline

marians:
Nick and I talked and I now understood the problem with the provider-specific implementation for cluster look-up. So we decided to release the command with the required `--provider` flag and create an issue to get the implementation changed and remove the need for the flag in the future.

marians (approved):

PR #628, opened 2021-12-21
Add in-cluster flag to template app command

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/20230

Currently `kubectl-gs template app` can only create App CRs for WC Apps. To support creating clusters using App CRs in a CAPI management cluster, I am adding a new flag `--in-cluster` which will install the app with `.kubeconfig.inCluster` set to `true` and adjust namespaces and names to suit our naming conventions.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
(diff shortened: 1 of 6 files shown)


MarcelMue (approved):
Makes sense to me - just reflecting more functionality present in the app CR here.

rossf7 (approved):

PR #635, opened 2022-01-17
Support templating Openstack clusters as App CRs

ljakimczuk:
### Description

The `kubectl gs template cluster --provider openstack` is now able to output clusters as App CRs. It does that by running two additional commands underneath, one for templating the [cluste-openstack](https://github.com/giantswarm/cluster-openstack) app and another for templating the [default-apps-openstack](https://github.com/giantswarm/default-apps-openstack) app.

In order to accomplished that, the below flags have been introduced:
* `--cluster-topology`, set to `false` by default, when used cluster will be templated as App CR,
* `--cluster-user-configmap`, the current command supports multiple flags for customizing cluster templating, though in the case of App CRs, the configuration must be provided in a form of `values.yaml`. We could try to translate the existing cluster flags to `values.yaml` before running the `template app` commands, and in fact I tried that at the beginning, though it quickly started to look messy and felt counterintuitive. Working with the `values.yaml` directly seems much more easier and flexible in configuration to the users, so I introduced this extra flag,
* `--default-apps-user-configmap`, the same reasoning as above applies here,
* `--cluster-app-version`, allows cluster app version selection,
* `--default-apps-app-version`, allows default apps app version selection.

### Examples

1. `$ kubectl gs template cluster --provider openstack --cluster-topology --organization giantswarm-2`

Output:
```yaml
---
kind: App
metadata:
  labels:
  name: ox0f8-cluster
  namespace: org-giantswarm-2
spec:
  catalog: control-plane-catalog
  config:
    configMap:
      name: ""
      namespace: ""
    secret:
      name: ""
      namespace: ""
  kubeConfig:
    context:
      name: ""
    inCluster: true
    secret:
      name: ""
      namespace: ""
  name: cluster-openstack
  namespace: org-giantswarm-2
  version: 0.1.0
---
kind: App
metadata:
  labels:
  name: ox0f8-default-apps
  namespace: org-giantswarm-2
spec:
  catalog: default
  config:
    configMap:
      name: ""
      namespace: ""
    secret:
      name: ""
      namespace: ""
  kubeConfig:
    context:
      name: ""
    inCluster: true
    secret:
      name: ""
      namespace: ""
  name: default-apps-openstack
  namespace: org-giantswarm-2
  version: 0.1.0
```

2. `$ kubectl gs template cluster --provider openstack --cluster-topology --organization giantswarm-2 --cluster-user-configmap /tmp/cm.yaml`

```yaml
---
apiVersion: v1
data:
  values: |
    clusterDescription: "My test cluster"
    organization: "giantswarm-2"
    cloudConfig: cloud-config
    cloudName: openstack

    kubernetesVersion: 1.20.9
    releaseVersion: 20.0.0-alpha1

    nodeClasses:
    - name: small
      machineFlavor: n1.small
      diskSize: 20

    controlPlane:
      class: default
      replicas: 3

    nodePools:
    - name: default
      class: default
      replicas: 3
kind: ConfigMap
metadata:
  creationTimestamp: null
  name: zi95g-cluster-userconfig
  namespace: org-giantswarm-2
---
kind: App
metadata:
  labels:
  name: zi95g-cluster
  namespace: org-giantswarm-2
spec:
  catalog: control-plane-catalog
  config:
    configMap:
      name: ""
      namespace: ""
    secret:
      name: ""
      namespace: ""
  kubeConfig:
    context:
      name: ""
    inCluster: true
    secret:
      name: ""
      namespace: ""
  name: cluster-openstack
  namespace: org-giantswarm-2
  userConfig:
    configMap:
      name: zi95g-cluster-userconfig
      namespace: org-giantswarm-2
  version: 0.1.0
---
kind: App
metadata:
  labels:
  name: zi95g-default-apps
  namespace: org-giantswarm-2
spec:
  catalog: default
  config:
    configMap:
      name: ""
      namespace: ""
    secret:
      name: ""
      namespace: ""
  kubeConfig:
    context:
      name: ""
    inCluster: true
    secret:
      name: ""
      namespace: ""
  name: default-apps-openstack
  namespace: org-giantswarm-2
  version: 0.1.0
```

3. `$ kubectl gs template cluster --provider openstack --organization giantswarm-2 --release 20.0.0-alpha1 --node-machine-flavor=n1.small` (some output has been cut for brevity)

```yaml
---
apiVersion: cluster.x-k8s.io/v1alpha4
kind: Cluster
metadata:
  annotations:
    cluster.giantswarm.io/description: ""
  labels:
    cluster.x-k8s.io/cluster-name: kz4j9
    giantswarm.io/cluster: kz4j9
    giantswarm.io/organization: giantswarm-2
    release.giantswarm.io/version: 20.0.0-alpha1
  name: kz4j9
  namespace: org-giantswarm-2
spec:
  controlPlaneEndpoint:
    host: ""
    port: 0
  controlPlaneRef:
    apiVersion: controlplane.cluster.x-k8s.io/v1alpha4
    kind: KubeadmControlPlane
    name: kz4j9
    namespace: org-giantswarm-2
  infrastructureRef:
    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
    kind: OpenStackCluster
    name: kz4j9
    namespace: org-giantswarm-2
status:
  infrastructureReady: false
---
...
---
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
kind: OpenStackMachineTemplate
metadata:
  labels:
    cluster.x-k8s.io/cluster-name: kz4j9
    giantswarm.io/cluster: kz4j9
    giantswarm.io/organization: giantswarm-2
    release.giantswarm.io/version: 20.0.0-alpha1
  name: kz4j9
  namespace: org-giantswarm-2
spec:
  template:
    spec:
      cloudName: openstack
      flavor: n1.small
      identityRef:
        kind: Secret
        name: cloud-config
      image: ubuntu-2004-kube-v1.20.9
      rootVolume: {}
```

### DIFF
--- a/cmd/template/app/runner.go
+++ b/cmd/template/app/runner.go
@@ -76,8 +76,8 @@ func (r *Runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		namespace = r.Flag.Cluster
 	}
 
-	if r.Flag.flagUserSecret != "" {
-		userConfigSecretData, err := key.ReadSecretYamlFromFile(afero.NewOsFs(), r.Flag.flagUserSecret)
+	if r.Flag.FlagUserSecret != "" {
+		userConfigSecretData, err := key.ReadSecretYamlFromFile(afero.NewOsFs(), r.Flag.FlagUserSecret)
 		if err != nil {
 			return microerror.Mask(err)
 		}
@@ -125,13 +125,13 @@ func (r *Runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		}
 	}
 
-	namespaceAnnotations, err := annotations.Parse(r.Flag.flagNamespaceConfigAnnotations)
+	namespaceAnnotations, err := annotations.Parse(r.Flag.FlagNamespaceConfigAnnotations)
 	if err != nil {
 		return microerror.Mask(err)
 	}
 	appConfig.NamespaceConfigAnnotations = namespaceAnnotations
 
-	namespaceLabels, err := labels.Parse(r.Flag.flagNamespaceConfigLabels)
+	namespaceLabels, err := labels.Parse(r.Flag.FlagNamespaceConfigLabels)
 	if err != nil {
 		return microerror.Mask(err)
 	}
--- a/cmd/template/app/runner.go
+++ b/cmd/template/app/runner.go
@@ -8,7 +8,6 @@ import (
 
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
-	"github.com/spf13/afero"
 	"github.com/spf13/cobra"
 	"sigs.k8s.io/yaml"
 
@@ -18,17 +17,17 @@ import (
 	templateapp "github.com/giantswarm/kubectl-gs/pkg/template/app"
 )
 
-type Runner struct {
-	Flag   *Flag
-	Logger micrologger.Logger
-	Stdout io.Writer
-	Stderr io.Writer
+type runner struct {
+	flag   *flag
+	logger micrologger.Logger
+	stderr io.Writer
+	stdout io.Writer
 }
 
-func (r *Runner) Run(cmd *cobra.Command, args []string) error {
+func (r *runner) Run(cmd *cobra.Command, args []string) error {
 	ctx := context.Background()
 
-	err := r.Flag.Validate()
+	err := r.flag.Validate()
 	if err != nil {
 		return microerror.Mask(err)
 	}
@@ -41,49 +40,44 @@ func (r *Runner) Run(cmd *cobra.Command, args []string) error {
 	return nil
 }
 
-func (r *Runner) run(ctx context.Context, cmd *cobra.Command, args []string) error {
+func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) error {
 	var userConfigConfigMapYaml []byte
 	var userConfigSecretYaml []byte
 	var err error
 
-	appName := r.Flag.AppName
+	appName := r.flag.AppName
 	if appName == "" {
-		appName = r.Flag.Name
+		appName = r.flag.Name
 	}
 
 	appConfig := templateapp.Config{
 		AppName:           appName,
-		Catalog:           r.Flag.Catalog,
-		Cluster:           r.Flag.Cluster,
-		DefaultingEnabled: r.Flag.DefaultingEnabled,
-		InCluster:         r.Flag.InCluster,
-		Name:              r.Flag.Name,
-		Namespace:         r.Flag.Namespace,
-		Version:           r.Flag.Version,
+		Catalog:           r.flag.Catalog,
+		Cluster:           r.flag.Cluster,
+		DefaultingEnabled: r.flag.DefaultingEnabled,
+		InCluster:         r.flag.InCluster,
+		Name:              r.flag.Name,
+		Namespace:         r.flag.Namespace,
+		Version:           r.flag.Version,
 	}
 
 	var assetName string
-	if r.Flag.InCluster {
+	if r.flag.InCluster {
 		assetName = key.GenerateAssetName(appName, "userconfig")
 	} else {
-		assetName = key.GenerateAssetName(appName, "userconfig", r.Flag.Cluster)
+		assetName = key.GenerateAssetName(appName, "userconfig", r.flag.Cluster)
 	}
 
 	var namespace string
-	if r.Flag.InCluster {
-		namespace = r.Flag.Namespace
+	if r.flag.InCluster {
+		namespace = r.flag.Namespace
 	} else {
-		namespace = r.Flag.Cluster
+		namespace = r.flag.Cluster
 	}
 
-	if r.Flag.FlagUserSecret != "" {
-		userConfigSecretData, err := key.ReadSecretYamlFromFile(afero.NewOsFs(), r.Flag.FlagUserSecret)
-		if err != nil {
-			return microerror.Mask(err)
-		}
-
-		secretConfig := templateapp.SecretConfig{
-			Data:      userConfigSecretData,
+	if r.flag.flagUserSecret != "" {
+		secretConfig := templateapp.UserConfig{
+			Path:      r.flag.flagUserSecret,
 			Name:      assetName,
 			Namespace: namespace,
 		}
@@ -99,20 +93,13 @@ func (r *Runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		}
 	}
 
-	if r.Flag.FlagUserConfigMap != "" {
-		var configMapData string
-		if r.Flag.FlagUserConfigMap != "" {
-			configMapData, err = key.ReadConfigMapYamlFromFile(afero.NewOsFs(), r.Flag.FlagUserConfigMap)
-			if err != nil {
-				return microerror.Mask(err)
-			}
-		}
-
-		configMapConfig := templateapp.ConfigMapConfig{
-			Data:      configMapData,
+	if r.flag.flagUserConfigMap != "" {
+		configMapConfig := templateapp.UserConfig{
+			Path:      r.flag.flagUserConfigMap,
 			Name:      assetName,
 			Namespace: namespace,
 		}
+
 		userConfigMap, err := templateapp.NewConfigMap(configMapConfig)
 		if err != nil {
 			return microerror.Mask(err)
@@ -125,13 +112,13 @@ func (r *Runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		}
 	}
 
-	namespaceAnnotations, err := annotations.Parse(r.Flag.FlagNamespaceConfigAnnotations)
+	namespaceAnnotations, err := annotations.Parse(r.flag.flagNamespaceConfigAnnotations)
 	if err != nil {
 		return microerror.Mask(err)
 	}
 	appConfig.NamespaceConfigAnnotations = namespaceAnnotations
 
-	namespaceLabels, err := labels.Parse(r.Flag.FlagNamespaceConfigLabels)
+	namespaceLabels, err := labels.Parse(r.flag.flagNamespaceConfigLabels)
 	if err != nil {
 		return microerror.Mask(err)
 	}
@@ -142,13 +129,7 @@ func (r *Runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		return microerror.Mask(err)
 	}
 
-	type AppCROutput struct {
-		AppCR               string
-		UserConfigSecret    string
-		UserConfigConfigMap string
-	}
-
-	appCROutput := AppCROutput{
+	appCROutput := templateapp.AppCROutput{
 		AppCR:               string(appCRYaml),
 		UserConfigConfigMap: string(userConfigConfigMapYaml),
 		UserConfigSecret:    string(userConfigSecretYaml),
--- a/cmd/template/app/runner.go
+++ b/cmd/template/app/runner.go
@@ -18,17 +18,17 @@ import (
 	templateapp "github.com/giantswarm/kubectl-gs/pkg/template/app"
 )
 
-type runner struct {
-	flag   *flag
-	logger micrologger.Logger
-	stdout io.Writer
-	stderr io.Writer
+type Runner struct {
+	Flag   *Flag
+	Logger micrologger.Logger
+	Stdout io.Writer
+	Stderr io.Writer
 }
 
-func (r *runner) Run(cmd *cobra.Command, args []string) error {
+func (r *Runner) Run(cmd *cobra.Command, args []string) error {
 	ctx := context.Background()
 
-	err := r.flag.Validate()
+	err := r.Flag.Validate()
 	if err != nil {
 		return microerror.Mask(err)
 	}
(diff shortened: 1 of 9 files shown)


cmd/template/cluster/provider/capo.go
```diff
@@ -65,3 +66,40 @@ func WriteOpenStackTemplate(ctx context.Context, client k8sclient.Interface, out
...
+func WriteOpenStackTemplateAppCR(ctx context.Context, runner *templateapp.Runner, config ClusterCRsConfig) error {
+	var err error
+
+	clusterAppFlags := &templateapp.Flag{
+		AppName:           config.Name,
+		Catalog:           "control-plane-catalog",
+		InCluster:         true,
+		Name:              "cluster-openstack",
+		Namespace:         fmt.Sprintf("org-%s", config.Organization),
+		Version:           "0.1.0",
```
MarcelMue, line 79: This should most likely be configurable.

ljakimczuk: The`version` is hard coded just for a start and I will add a flag to configure it. And did you mean the other fields as well?

ljakimczuk: Sure, got it. I must also add another `user-configmap` for the app of apps

cmd/template/app/runner.go
```diff
@@ -18,17 +18,17 @@ import (
 	templateapp "github.com/giantswarm/kubectl-gs/pkg/template/app"
 )
 
-type runner struct {
-	flag   *flag
-	logger micrologger.Logger
-	stdout io.Writer
-	stderr io.Writer
+type Runner struct {
```
rossf7, line 21: Rather than exporting the runner I'd prefer more of the logic is moved to `pkg/template/app`.
Can you take a look and see how hard that would be?

Reason being that multiple teams need to work on `kubectl gs` so it's good to keep the code as consistent as possible. 

Also me may need to support more types of apps in future. e.g. an app of apps for the security stack.

cmd/template/cluster/flag.go
```diff
@@ -15,12 +15,19 @@ import (
 
 const (
 	flagProvider = "provider"
+	flagTplType  = "template-type"
```
rossf7, line 18:
```suggestion
	flagTemplateType  = "template-type"
```

Flags are not that long so I think its better to align the constants with the full name.

cmd/template/cluster/flag.go
```diff
@@ -15,12 +15,19 @@ import (
...
 	// AWS only.
 	flagAWSExternalSNAT       = "external-snat"
 	flagAWSEKS                = "aws-eks"
 	flagAWSControlPlaneSubnet = "control-plane-subnet"
 
+	// Cluster App only.
+	flagClusterAppVersion     = "cluster-app-version"
+	flagDefaultAppsAppVersion = "default-apps-app-version"
+	flagClusterConfigMap      = "cluster-user-configmap"
+	flagDefaultAppsConfigMap  = "default-apps-user-configmap"
```
rossf7, line 29:
```suggestion
	flagClusterUserConfigMap      = "cluster-user-configmap"
	flagDefaultAppsUserConfigMap  = "default-apps-user-configmap"
```

cmd/template/cluster/flag.go
```diff
@@ -93,6 +109,7 @@ type flag struct {
 
 func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.Provider, flagProvider, "", "Installation infrastructure provider.")
+	cmd.Flags().StringVar(&f.TplType, flagTplType, "raw", "Type of the template. Could be one of: raw, appcr. If application, cluster will be templated as an App CR. (default raw)")
```
rossf7, line 112:
```suggestion
	cmd.Flags().StringVar(&f.TplType, flagTplType, "raw", "Type of the template. Could be one of: raw, app. If application, cluster will be templated as an App CR. (default raw)")
```

We use apps in our docs already. Let's see what sig UX says later 😅 but I prefer just `app`.

cmd/template/cluster/flag.go
ljakimczuk: Makes sense, done it

cmd/template/cluster/flag.go
ljakimczuk: I agree `app` looks better 😄

cmd/template/app/runner.go
ljakimczuk: @rossf7  Agree that exporting apps introduced some inconsistency, it may hurt aesthetics 😄 I tried to re-organize the code a bit, to move some parts to the `package/template/app` and thus avoid that much code repetition, though I failed to fully accomplish the goal. At the moment the `pkg/template/app` provides some structs and functions, we relay on the returned values to modify instances of these structs in the `cmd/template/app`, that's why a good dose of code lives in the `cmd/template/app` and the `cmd/template/cluster/provider/capo`. I believe if we want to move more of it into the `package/template/app` we will need to make classes out of these structs, and make the functions to be their methods, but I am not sure if that's not too big of a change.

cmd/template/cluster/flag.go
```diff
@@ -93,6 +109,7 @@ type flag struct {
 
 func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.Provider, flagProvider, "", "Installation infrastructure provider.")
+	cmd.Flags().StringVar(&f.TemplateType, flagTemplateType, "raw", "Type of the template. Could be one of: raw, app. If application, cluster will be templated as an App CR. (default raw)")
```
rossf7, line 112:
```suggestion
	cmd.Flags().StringVar(&f.TemplateType, flagTemplateType, "raw", "Type of the template. Could be one of: raw, app. If app, cluster will be templated as an App CR. (default raw)")
```

cmd/template/cluster/provider/capo.go
```diff
@@ -65,3 +70,96 @@ func WriteOpenStackTemplate(ctx context.Context, client k8sclient.Interface, out
...
+		Catalog:   "control-plane-catalog",
+		InCluster: true,
+		Name:      "cluster-openstack",
+		Namespace: fmt.Sprintf("org-%s", config.Organization),
+		Version:   config.ClusterAppVersion,
+	}
+
+	defaultAppsAppConfig := templateapp.Config{
+		AppName:   "openstack-default-apps",
+		Catalog:   "control-plane-catalog",
```
rossf7, line 92:
```suggestion
		Catalog:   "default",
```

The app is only pushed to the default catalog right now. Since that's where the child apps are.

You can check this is in the Circle [config](https://github.com/giantswarm/default-apps-openstack/blob/3a994712a58b2183737201e85cbcbabee8498321/.circleci/config.yml#L12)

cmd/template/cluster/provider/common.go
```diff
@@ -51,6 +51,11 @@ type ClusterCRsConfig struct {
 	Labels         map[string]string
 	Namespace      string
 	PodsCIDR       string
+
+	ClusterAppUserConfigMap     string
+	DefaultAppsAppUserConfigMap string
+	ClusterAppVersion           string
+	DefaultAppsAppVersion       string
```
rossf7, line 58:
```suggestion
        // App settings.
	ClusterAppUserConfigMap     string
	ClusterAppVersion           string
	DefaultAppsAppUserConfigMap string
	DefaultAppsAppVersion       string
```

cmd/template/cluster/runner.go
```diff
@@ -78,6 +78,11 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 			RootVolumeDiskSize:   r.flag.OpenStack.RootVolumeDiskSize,
 			RootVolumeSourceType: r.flag.OpenStack.RootVolumeSourceType,
 			RootVolumeSourceUUID: r.flag.OpenStack.RootVolumeSourceUUID,
+
+			ClusterAppVersion:           r.flag.ClusterApp.ClusterAppVersion,
+			DefaultAppsAppVersion:       r.flag.ClusterApp.DefaultAppsAppVersion,
+			ClusterAppUserConfigMap:     r.flag.ClusterApp.ClusterUserConfigMap,
+			DefaultAppsAppUserConfigMap: r.flag.ClusterApp.DefaultAppsUserConfigMap,
```
rossf7, line 85:
```suggestion
			ClusterAppVersion:           r.flag.ClusterApp.ClusterAppVersion,
			ClusterAppUserConfigMap:     r.flag.ClusterApp.ClusterUserConfigMap,
			DefaultAppsAppVersion:       r.flag.ClusterApp.DefaultAppsAppVersion,
			DefaultAppsAppUserConfigMap: r.flag.ClusterApp.DefaultAppsUserConfigMap,
```

cmd/template/cluster/provider/capo.go
```diff
@@ -65,3 +70,96 @@ func WriteOpenStackTemplate(ctx context.Context, client k8sclient.Interface, out
...
+
+func WriteOpenStackTemplateAppCR(ctx context.Context, config ClusterCRsConfig) error {
+	var userConfigConfigMapYaml []byte
+	var err error
+
+	clusterAppOutput := templateapp.AppCROutput{}
+	defaultAppsAppOutput := templateapp.AppCROutput{}
+
+	clusterAppConfig := templateapp.Config{
+		AppName:   config.Name,
```
rossf7, line 82:
```suggestion
		AppName:   fmt.Sprintf("%s-cluster", config.Name),
```

cmd/template/cluster/provider/capo.go
```diff
@@ -65,3 +70,96 @@ func WriteOpenStackTemplate(ctx context.Context, client k8sclient.Interface, out
...
+		Catalog:   "control-plane-catalog",
+		InCluster: true,
+		Name:      "cluster-openstack",
+		Namespace: fmt.Sprintf("org-%s", config.Organization),
+		Version:   config.ClusterAppVersion,
+	}
+
+	defaultAppsAppConfig := templateapp.Config{
+		AppName:   "openstack-default-apps",
```
rossf7, line 91:
```suggestion
		AppName:   fmt.Sprintf("%s-default-apps", config.Name),
```

cmd/template/app/runner.go
rossf7: Thx! The structure looks a lot better ❤️ 

The main thing was I didn't want the runner to be exported. I think this is fine for now we can always revisit when we add support for more providers.

rossf7 (commented):
@ljakimczuk This is looking good but CI isn't happy. Can you check and appease the Circle gods? :)

cmd/template/app/runner.go
ljakimczuk: Perfect, good to hear that, I made the suggested changes and fixed import so that CI's happy.

cmd/template/cluster/provider/capo.go
ljakimczuk: Ok, the fields now have their own flags

cmd/template/cluster/flag.go
```diff
@@ -14,13 +14,20 @@ import (
...
 	// AWS only.
 	flagAWSExternalSNAT       = "external-snat"
 	flagAWSEKS                = "aws-eks"
 	flagAWSControlPlaneSubnet = "control-plane-subnet"
 
+	// Cluster App only.
+	flagClusterAppVersion        = "cluster-app-version"
+	flagDefaultAppsAppVersion    = "default-apps-app-version"
+	flagClusterUserConfigMap     = "cluster-user-configmap"
+	flagDefaultAppsUserConfigMap = "default-apps-user-configmap"
```
rossf7, line 29: Nit but could you order the flags in the group in alpha order?

That looks tidier to me 🤓

rossf7 (commented):
Nice<3 this is looking good. Just some final nits on ordering.

We should get a review from Sig UX before merging but we can start with a review from the team.

We'll also need a docs PR 📚 https://docs.giantswarm.io/ui-api/kubectl-gs/template-cluster/

cmd/template/cluster/flag.go
ljakimczuk: Sure thing!

ljakimczuk:
> We'll also need a docs PR.

I remember about that, but let's maybe wait for that until we get more reviews, just in case we have more changes coming in. WDYT?

rossf7:
> maybe wait for that until we get more reviews, just in case we have more changes coming in. WDYT?

@ljakimczuk Yes that's fine 👍

cmd/template/cluster/flag.go
```diff
@@ -112,6 +129,12 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.OpenStack.RootVolumeSourceUUID, flagOpenStackRootVolumeSourceUUID, "", "Root volume source UUID (OpenStack only).")
 	cmd.Flags().StringVar(&f.OpenStack.NodeCIDR, flagOpenStackNodeCIDR, "", "CIDR used for the nodes.")
 
+	// OpenStack App only.
+	cmd.Flags().StringVar(&f.ClusterApp.ClusterAppVersion, flagClusterAppVersion, "0.1.0", "Cluster App version to be installed. (OpenStack App CR only).")
```
tfussell, line 133: I think we want to hide these flags for now, see https://github.com/giantswarm/kubectl-gs/pull/540#issuecomment-977888540 and https://github.com/giantswarm/roadmap/issues/605. WDYT @giantswarm/team-rocket?

tfussell (approved):

cmd/template/cluster/flag.go
anvddriesch: I would suggest this, too.

anvddriesch (commented):
Cool stuff!
Please list all the changes to the `CHANGELOG`.

rossf7 (approved):
LGTM when @anvddriesch and team 🌈 are happy

Agree it makes sense to hide the openstack flags for now.

cmd/template/cluster/flag.go
ljakimczuk: Ok, it's done

ljakimczuk:
@rossf7 @anvddriesch @tfussell done, flags are now hidden and the changlog entry is there, let me know if I should make it more detailed.

anvddriesch (approved):
LGTM

ljakimczuk:
@kopiczko not necessarily, it seemed more consistent to me to have this type, but it would work just as good without it. We can get rid of it or change this flag to something not suggesting multiple types, for example, `--as-app` or similar

kopiczko:
> can get rid of it or change this flag to something not suggesting multiple types, for example, `--as-app` or similar

I thought we could replace existing templates and ditch them altogether.

ljakimczuk:
I believe the original idea was to support both templates types, at least for some time since some users may be interested in templating raw resources, but maybe I got it wrong. @rossf7 WDYT?

kopiczko:
Asked @cornelius-keller

cornelius-keller:
I think currently it does not hurt and gives the customer the opportunity to create the clusters with the old method still, which might be useful in case there are problems with the new ones and we need to compare them to see what is different.

In the long run I agree with @kopiczko and it should disappear or be renamed, especially because customers can expect that the raw method templates  the same cluster as the app method only that it is the raw template and not wrapped into an app.

Maybe it will be useful to provide an option to get a raw template that is the same as included in the app, but that is a different story.

LGTM

kopiczko:
It would be nice to at least generate raw templates using `helm template` to make sure they are always on par.

ljakimczuk:
> It would be nice to at least generate raw templates using helm template to make sure they are always on par.

This may be misleading for those relying on the "old" version for the migration time being and as far as I understood that's indeed the case. But since `app` is mostly about adding `ClusterTopology`, @tfussell suggested we could change the flag to `--cluster-topology [true|false]`, @tfussell would you like to add something to that?

cmd/template/cluster/provider/capo.go
```diff
@@ -65,3 +71,97 @@ func WriteOpenStackTemplate(ctx context.Context, client k8sclient.Interface, out
...
+
+func WriteOpenStackTemplateAppCR(ctx context.Context, config ClusterCRsConfig) error {
+	var userConfigConfigMapYaml []byte
+	var err error
+
+	clusterAppOutput := templateapp.AppCROutput{}
+	defaultAppsAppOutput := templateapp.AppCROutput{}
+
+	clusterAppConfig := templateapp.Config{
+		AppName:   fmt.Sprintf("%s-cluster", config.Name),
```
rossf7, line 83: @ljakimczuk @tfussell I found a problem. We use `.Release.Name` in the cluster helm chart which turns out badly :(

```
ross3-cluster   Provisioned   4m37s   v1.20.9
```

Should we remove the `-cluster` suffix here or in the helm chart?

ljakimczuk: I guess we could remove it from here, though App name may then be vague to the user, cause in some cases it would be auto-generated

marians:
Haven't read through the entire discussion.

One minor thing that stands out to me is the duplication of the term `template` in the flag `--template-type`. If we need that flag, could it simply be `--type` ?

ljakimczuk:
> One minor thing that stands out to me is the duplication of the term template in the flag --template-type. If we need that flag, could it simply be --type

I duplicated term `template` to emphasize the fact it's about the template type. I was afraid purpose of the `--type` alone may be vague having all the providers and different resources generated. 

@tfussell has already proposed to change this flag to something else, namely `--cluster-topology [true|false]`, that more closely describe the purpose of having an app for cluster, what do you think @marians ?

When we have an agreement I will push the changes.

marians:
With a boolean flag, the default (when the flag is not used) should always be `false`. So `--cluster-topology false` would be the default. I'd like to make sure that the default makes sense.

What sort of output will that (`--cluster-topology false`) generate then, and what will  `--cluster-topology` or `--cluster-topology true` generate instead?

ljakimczuk:
@marians The `--cluster-topology false` in  this scenario would behave the same way the `--template-type raw` (or lack of it) behaves. The `cluster-topology true` would behave the same way the `--template-type app` behaves right now.

marians (approved):

PR #638, opened 2022-01-21
Enable cluster topology for OpenStack by default and update cluster-openstack default version to 0.3.0

tfussell:
The latest version of cluster-openstack is 0.3.0.

In the next `kubectl-gs` release we will get rid of the old non-cluster topology cluster templating entirely. During the transition it should be the default.

### DIFF
--- a/cmd/template/cluster/flag.go
+++ b/cmd/template/cluster/flag.go
@@ -130,7 +130,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 
 	// OpenStack App only.
 	cmd.Flags().StringVar(&f.ClusterApp.ClusterUserConfigMap, flagClusterUserConfigMap, "", "Path to the user values configmap YAML file for Cluster App (OpenStack App CR only).")
-	cmd.Flags().StringVar(&f.ClusterApp.ClusterAppVersion, flagClusterAppVersion, "0.1.0", "Cluster App version to be installed. (OpenStack App CR only).")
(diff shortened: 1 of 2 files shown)


anvddriesch (approved):

rossf7 (approved):

cmd/template/cluster/flag.go
```diff
@@ -130,7 +130,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 
 	// OpenStack App only.
 	cmd.Flags().StringVar(&f.ClusterApp.ClusterUserConfigMap, flagClusterUserConfigMap, "", "Path to the user values configmap YAML file for Cluster App (OpenStack App CR only).")
-	cmd.Flags().StringVar(&f.ClusterApp.ClusterAppVersion, flagClusterAppVersion, "0.1.0", "Cluster App version to be installed. (OpenStack App CR only).")
+	cmd.Flags().StringVar(&f.ClusterApp.ClusterAppVersion, flagClusterAppVersion, "0.2.1", "Cluster App version to be installed. (OpenStack App CR only).")
```
kopiczko, line 133:
```suggestion
	cmd.Flags().StringVar(&f.ClusterApp.ClusterAppVersion, flagClusterAppVersion, "0.2.2", "Cluster App version to be installed. (OpenStack App CR only).")
```

marians: If 0.2.2 is the right version, please update the PR title and description.

tfussell: Thanks, yes, I was holding off as I was still deciding what the correct default version should be. I will update the title and description if I change it.

tfussell:
Made another flag change and updated the version to match with the latest release.

MarcelMue (approved):

PR #639, opened 2022-01-21
[WIP] Add `organization` flag for templating App CR in org namespace

ljakimczuk:
### Description

This PR is part of the ongoing migration towards handling App CRs in the org namespace instead of cluster namespaces. 

Related issue: https://github.com/giantswarm/giantswarm/issues/20443
Related PRs: https://github.com/giantswarm/cluster-apps-operator/pull/148

This PR adds `organization` flag to the `template app` command. This flag is now optional, and thus is not validated, to support old behavior of templating App CR into cluster namespace.

Upon passing this new flag to the command, App CR will be templated for the org-namespace.

#### Without `organization` flag

1. `kubectl gs template app --name hello-world --version 0.1.0 --namespace default --cluster ab23c --catalog giantswarm`

```yaml
---
kind: App
metadata:
  name: hello-world
  namespace: ab23c
spec:
  catalog: giantswarm
  kubeConfig:
    inCluster: false
  name: hello-world
  namespace: default
  version: 0.1.0
```

2. `kubectl gs template app --name hello-world --version 0.1.0 --namespace default --cluster ab23c --catalog giantswarm --user-configmap /tmp/userconfig.yaml`

```yaml
---
apiVersion: v1
data:
  values: |
    doesThisConfigMakeSense:
      notReally: true
kind: ConfigMap
metadata:
  creationTimestamp: null
  name: hello-world-userconfig-ab23c
  namespace: ab23c
---
kind: App
metadata:
  name: hello-world
  namespace: ab23c
spec:
  catalog: giantswarm
  kubeConfig:
    inCluster: false
  name: hello-world
  namespace: default
  userConfig:
    configMap:
      name: hello-world-userconfig-ab23c
      namespace: ab23c
  version: 0.1.0
```

#### With `organization` flag

1. `kubectl gs template app --name hello-world --version 0.1.0 --namespace default --cluster ab23c --catalog giantswarm --organization giantswarm`

```yaml
---
kind: App
metadata:
  labels:
    giantswarm.io/cluster: ab23c
  name: hello-world
  namespace: org-giantswarm
spec:
  catalog: giantswarm
  kubeConfig:
    inCluster: false
  name: hello-world
  namespace: default
  version: 0.1.0
```

2. `kubectl template app --name hello-world --version 0.1.0 --namespace default --cluster ab23c --catalog giantswarm --organization giantswarm --user-configmap /tmp/userconfig.yaml`

```yaml
---
apiVersion: v1
data:
  values: |
    doesThisConfigMakeSense:
      notReally: true
kind: ConfigMap
metadata:
  creationTimestamp: null
  name: hello-world-userconfig-ab23c
  namespace: org-giantswarm
---
kind: App
metadata:
  labels:
    giantswarm.io/cluster: ab23c
  name: hello-world
  namespace: org-giantswarm
spec:
  catalog: giantswarm
  kubeConfig:
    inCluster: false
  name: hello-world
  namespace: default
  userConfig:
    configMap:
      name: hello-world-userconfig-ab23c
      namespace: org-giantswarm
  version: 0.1.0
```

### DIFF
--- a/cmd/template/app/flag.go
+++ b/cmd/template/app/flag.go
@@ -20,6 +20,7 @@ const (
 	flagNamespace                  = "namespace"
 	flagNamespaceConfigAnnotations = "namespace-annotations"
 	flagNamespaceConfigLabels      = "namespace-labels"
+	flagOrganization               = "organization"
 	flagUserConfigMap              = "user-configmap"
 	flagUserSecret                 = "user-secret"
 	flagVersion                    = "version"
@@ -33,6 +34,7 @@ type flag struct {
 	InCluster                      bool
 	Name                           string
 	Namespace                      string
+	Organization                   string
 	Version                        string
 	flagNamespaceConfigAnnotations []string
 	flagNamespaceConfigLabels      []string
@@ -46,6 +48,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.Name, flagName, "", "Name of the app in the Catalog.")
 	cmd.Flags().StringVar(&f.Namespace, flagNamespace, "", "Namespace where the app will be deployed.")
 	cmd.Flags().StringVar(&f.Cluster, flagCluster, "", "Name of the cluster the app will be deployed to.")
+	cmd.Flags().StringVar(&f.Organization, flagOrganization, "", "Workload cluster organization.")
 	cmd.Flags().BoolVar(&f.DefaultingEnabled, flagDefaultingEnabled, true, "Don't template fields that will be defaulted.")
 	cmd.Flags().BoolVar(&f.InCluster, flagInCluster, false, fmt.Sprintf("Deploy the app in the current management cluster rather than in a workload cluster. If this is set, --%s will be ignored.", flagCluster))
 	cmd.Flags().StringVar(&f.flagUserConfigMap, flagUserConfigMap, "", "Path to the user values configmap YAML file.")
@@ -68,6 +71,12 @@ func (f *flag) Validate() error {
 	if !f.InCluster && f.Cluster == "" {
 		return microerror.Maskf(invalidFlagError, "--%s must not be empty", flagCluster)
 	}
+	// Once we fully migrate to org namespace, checking for
+	// the `organization` flag should be mandatory.
+	//
+	/*if !f.InCluster && f.Organization == "" {
+		return microerror.Maskf(invalidFlagError, "--%s must not be empty", flagOrganization)
+	}*/
 	if f.Version == "" {
 		return microerror.Maskf(invalidFlagError, "--%s must not be empty", flagVersion)
 	}
--- a/cmd/template/app/runner.go
+++ b/cmd/template/app/runner.go
@@ -2,8 +2,10 @@ package app
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"os"
+	"strings"
 	"text/template"
 
 	"github.com/giantswarm/microerror"
@@ -50,6 +52,10 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		appName = r.flag.Name
 	}
 
+	// Since organization may be provided in a mixed-cases form,
+	// make sure it is lowercased before using.
+	organization := strings.ToLower(r.flag.Organization)
+
 	appConfig := templateapp.Config{
 		AppName:           appName,
 		Catalog:           r.flag.Catalog,
@@ -58,6 +64,7 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		InCluster:         r.flag.InCluster,
 		Name:              r.flag.Name,
 		Namespace:         r.flag.Namespace,
+		Organization:      organization,
 		Version:           r.flag.Version,
 	}
 
@@ -68,9 +75,13 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		assetName = key.GenerateAssetName(appName, "userconfig", r.flag.Cluster)
 	}
 
+	// If organization is passed to the command use it as the indicator for
+	// templating App CR in org namespace.
 	var namespace string
 	if r.flag.InCluster {
 		namespace = r.flag.Namespace
+	} else if appConfig.Organization != "" {
+		namespace = fmt.Sprintf("org-%s", appConfig.Organization)
 	} else {
 		namespace = r.flag.Cluster
 	}
--- a/pkg/template/app/app.go
+++ b/pkg/template/app/app.go
@@ -1,6 +1,8 @@
 package app
 
 import (
+	"fmt"
+
 	"github.com/giantswarm/apiextensions/v3/pkg/apis/application/v1alpha1"
 	applicationv1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/application/v1alpha1"
 	"github.com/giantswarm/microerror"
@@ -9,6 +11,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/yaml"
 
+	"github.com/giantswarm/k8smetadata/pkg/label"
 	"github.com/giantswarm/kubectl-gs/internal/key"
 )
 
@@ -24,6 +27,7 @@ type Config struct {
 	NamespaceConfigLabels      map[string]string
 	UserConfigConfigMapName    string
 	UserConfigSecretName       string
+	Organization               string
 	Version                    string
 }
 
@@ -41,10 +45,16 @@ type AppCROutput struct {
 
 func NewAppCR(config Config) ([]byte, error) {
 	userConfig := applicationv1alpha1.AppSpecUserConfig{}
+	appLabels := map[string]string{}
 
+	// When templating App CR for an organization namespace add
+	// `giantswarm.io/cluster` label for cluster selection.
 	var namespace string
 	if config.InCluster {
 		namespace = config.Namespace
+	} else if config.Organization != "" {
+		namespace = fmt.Sprintf("org-%s", config.Organization)
+		appLabels[label.Cluster] = config.Cluster
 	} else {
 		namespace = config.Cluster
 	}
@@ -71,6 +81,7 @@ func NewAppCR(config Config) ([]byte, error) {
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      config.AppName,
 			Namespace: namespace,
+			Labels:    appLabels,
 		},
 		Spec: applicationv1alpha1.AppSpec{
 			Catalog:   config.Catalog,
--- a/pkg/template/app/app_test.go
+++ b/pkg/template/app/app_test.go
@@ -82,6 +82,20 @@ func Test_NewAppCR(t *testing.T) {
 			},
 			expectedGoldenFile: "app_override_app_name_yaml_output.golden",
 		},
+		{
+			name: "case 5: flawless flow for organization",
+			config: Config{
+				AppName:           "nginx-ingress-controller-app",
+				Catalog:           "giantswarm",
+				Cluster:           "eggs2",
+				DefaultingEnabled: true,
+				Name:              "nginx-ingress-controller-app",
+				Namespace:         "kube-system",
+				Organization:      "giantswarm",
+				Version:           "1.17.0",
+			},
+			expectedGoldenFile: "app_flawless_flow_organization_yaml_output.golden",
+		},
 	}
 
 	for i, tc := range testCases {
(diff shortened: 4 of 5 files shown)


cmd/template/app/flag.go
```diff
@@ -68,6 +71,12 @@ func (f *flag) Validate() error {
 	if !f.InCluster && f.Cluster == "" {
 		return microerror.Maskf(invalidFlagError, "--%s must not be empty", flagCluster)
 	}
+	// Once we fully migrate to org namespace, checking for
+	// the `organization` flag should be mandatory.
+	//
+	/*if !f.InCluster && f.Organization == "" {
+		return microerror.Maskf(invalidFlagError, "--%s must not be empty", flagOrganization)
+	}*/
```
rossf7, line 79: The problem is this migration will take many months. I think maybe we could have some kind of feature detection instead of flag validation.

cmd/template/app/runner.go
```diff
@@ -50,6 +52,10 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		appName = r.flag.Name
 	}
 
+	// Since organization may be provided in a mixed-cases form,
+	// make sure it is lowercased before using.
```
rossf7, line 56:
```suggestion
	// Since organization may be provided in a mixed-cased form,
	// make sure it is lowercased.
```

Nice 👍

rossf7 (commented):
Looks good I'm just not sure how we can validate the org flag.

However if we change the validation logic in https://github.com/giantswarm/giantswarm/issues/20448 we can check in app-admission-controller if the namespace and label combination is correct.

cmd/template/app/flag.go
ljakimczuk: Sure, I believe we do not need this basic flag validation, I put it there just as an example of what we will eventually end up with. But for now we could treat this flag as a feature detection mechanism, since it presence does reflect of how the App CR is templated. Or we could add a boolean flag like `--capo` and then upon user providing it, demand and validate the `organization`. WDYT @rossf7 ?

rossf7: I would remove it for now but mention in the PR that we can't validate the flag.

> But for now we could treat this flag as a feature detection mechanism, since it presence does reflect of how the App CR is templated.

Yes lets use it as a feature flag for now as we only need it for CAPO.

> Or we could add a boolean flag like --capo and then upon user providing it, demand and validate the organization

We try to keep the app platform decoupled from the provider. I think our best option is to extend app-admission-controller and give the user a useful error if the cluster label is missing. WDYT?

ljakimczuk: I actually very like the idea of extending the app-admission-controller

ljakimczuk: Let's put it on hold then, I will submit a PR to app-admission-controller and link it here, so that we can then advance with this PR.

ljakimczuk: WDYT @rossf7 ?

rossf7: Good call and there is also https://github.com/giantswarm/giantswarm/issues/20448 for that.

We can then drop the version label from the `default-apps-openstack` chart 🔥

ljakimczuk: Ok, we got it: https://github.com/giantswarm/app/pull/206 and https://github.com/giantswarm/app-admission-controller/pull/222

rossf7: @ljakimczuk We're going to need some more changes to remove usage of the cluster namespace.

There is also the flag `--defaulting-enabled` when its false it renders more fields.

```
k gs template app \
  --catalog=giantswarm \
  --cluster=eggs2 \
  --name=nginx-ingress-controller-app \
  --namespace=kube-system \
  --version=1.17.0 \
  --defaulting-enabled=false \
  --organization=test
---
kind: App
metadata:
  labels:
    app-operator.giantswarm.io/version: 1.0.0
  name: nginx-ingress-controller-app
  namespace: org-test
spec:
  catalog: giantswarm
  config:
    configMap:
      name: eggs2-cluster-values
      namespace: eggs2
    secret:
      name: ""
      namespace: ""
  kubeConfig:
    context:
      name: eggs2-kubeconfig
    inCluster: false
    secret:
      name: eggs2-kubeconfig
      namespace: eggs2
  name: nginx-ingress-controller-app
  namespace: kube-system
  version: 1.17.0
```

The cluster values CM and the kubeconfig secret need to be in the org namespace. We should also check the defaulting logic in app-admission-controller.

Can you also check the output if the user sets `--user-configmap` or `--user-secret`?

ljakimczuk: @rossf7 `--user-configmap` example is already given in the PR description, I will also add there another example with `--user-secret`, but indeed I missed the `defaulting-enabled` and  there seems to be something wrong with it

rossf7: > --user-configmap example is already given in the PR description

@ljakimczuk Ah my bad I missed that 👍

ljakimczuk: @rossf7 my bad, there is still one issue there

pkg/template/app/app.go
```diff
@@ -41,25 +46,38 @@ type AppCROutput struct {
...
-	var namespace string
+	appLabels := map[string]string{}
+
+	// Accomodating all the label cases here:
+	// 1. In-cluster Apps get unique label
+	// 2. Org-namespaced Apps get cluster label
+	// 3. Cluster-namespaced Apps with defaulting disabled gets default version label
+	// 4. Cluster-namespaced Apps with defaulting enabled gets nothing
+	var crNamespace string
 	if config.InCluster {
```
ljakimczuk, line 57: Moved labels logic here, to have a single place to decide which ones to set. It seems cleaner to me, WDYT?

cmd/template/app/flag.go
ljakimczuk: @rossf7 should be good now, WDYT?

ljakimczuk: In the meantime I will check the defaulting in app admission controller

pkg/template/app/app.go
```diff
@@ -41,25 +46,31 @@ type AppCROutput struct {
 
 func NewAppCR(config Config) ([]byte, error) {
 	userConfig := applicationv1alpha1.AppSpecUserConfig{}
+	appLabels := map[string]string{}
 
-	var namespace string
+	// When templating App CR for an organization namespace add
+	// `giantswarm.io/cluster` label for cluster selection.
+	var crNamespace string
```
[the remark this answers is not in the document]
rossf7: Good to clarify that. We sometimes use target namespace to refer to `.spec.namespace` which I think is useful.

This is the same naming as in the [HelmRelease](https://fluxcd.io/docs/components/helm/helmreleases/) CRD. I think in a future version of the App CRD we should rename it.

pkg/template/app/app.go
```diff
@@ -41,25 +46,38 @@ type AppCROutput struct {
...
 	if config.InCluster {
-		namespace = config.Namespace
+		crNamespace = config.Namespace
+		appLabels[label.AppOperatorVersion] = "0.0.0"
+	} else if config.Organization != "" {
+		crNamespace = fmt.Sprintf("org-%s", config.Organization)
+		appLabels[label.Cluster] = config.Cluster
+	} else if !config.DefaultingEnabled {
+		appLabels[label.AppOperatorVersion] = "1.0.0"
```
rossf7, line 64: There is some technical debt here :( `1.0.0` was used for app CRs using helm 2 and is now no longer in use.

If you apply this does app-admission-controller update the value to the correct version for the cluster?

cmd/template/app/flag.go
```diff
@@ -46,6 +48,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.Name, flagName, "", "Name of the app in the Catalog.")
 	cmd.Flags().StringVar(&f.Namespace, flagNamespace, "", "Namespace where the app will be deployed.")
 	cmd.Flags().StringVar(&f.Cluster, flagCluster, "", "Name of the cluster the app will be deployed to.")
+	cmd.Flags().StringVar(&f.Organization, flagOrganization, "", "Workload cluster organization.")
```
rossf7, line 51: I have a concern about the flag. As right now this will only work for CAPO clusters.

IIRC we hid the CAPO specific flags for `template cluster`. Should we do the same here until more providers are supported?

ljakimczuk: Good point, this flag is indeed going to be ignored for non-CAPO clusters, so hiding it makes total sense.

pkg/template/app/app.go
ljakimczuk: Let me check that, I haven't tried to apply it live yet

pkg/template/app/app.go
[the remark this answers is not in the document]
ljakimczuk: yeah, target namespace would be more obvious 😄

pkg/template/app/app.go
ljakimczuk: @rossf7 update from my side, when I submitted app with `1.0.0` label it was updated to the correct value. If that's the case I guess we can get rid of this part of labeling logic, WDYT?

rossf7: @ljakimczuk Thanks for checking and yes let's remove the version label here so its defaulted.

We don't want customers adding it to their gitops repo as its an implementation detail. There is an `appversionupdate` resource in cluster-operator that ensures the label value is set correctly.

cmd/template/app/flag.go
ljakimczuk: Ok, flag is hidden now

pkg/template/app/app.go
ljakimczuk: Done, we no longer set the `1.0.0` version

pkg/template/app/app.go
```diff
@@ -9,6 +11,8 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/yaml"
 
+	"github.com/giantswarm/k8smetadata/pkg/label"
```
MarcelMue, line 14: Seems to me like imports are all kind of out of order here

ljakimczuk: Does it look correct now?

pkg/template/app/app.go
```diff
@@ -1,15 +1,17 @@
...
 	applicationv1alpha1 "github.com/giantswarm/apiextensions/v3/pkg/apis/application/v1alpha1"
+	"github.com/giantswarm/k8smetadata/pkg/label"
+	"github.com/giantswarm/kubectl-gs/internal/key"
 	"github.com/giantswarm/microerror"
 	"github.com/spf13/afero"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/yaml"
-
-	"github.com/giantswarm/kubectl-gs/internal/key"
```
MarcelMue, line 12: This should remain in its own block as it's project internal - LGTM apart from that.

MarcelMue (approved):

rossf7 (approved):
LGTM just needs a changelog :)

cmd/template/app/flag.go
```diff
@@ -46,13 +48,16 @@ func (f *flag) Init(cmd *cobra.Command) {
...
 	cmd.Flags().BoolVar(&f.DefaultingEnabled, flagDefaultingEnabled, true, "Don't template fields that will be defaulted.")
 	cmd.Flags().BoolVar(&f.InCluster, flagInCluster, false, fmt.Sprintf("Deploy the app in the current management cluster rather than in a workload cluster. If this is set, --%s will be ignored.", flagCluster))
 	cmd.Flags().StringVar(&f.flagUserConfigMap, flagUserConfigMap, "", "Path to the user values configmap YAML file.")
 	cmd.Flags().StringVar(&f.flagUserSecret, flagUserSecret, "", "Path to the user secrets YAML file.")
 	cmd.Flags().StringVar(&f.Version, flagVersion, "", "App version to be installed.")
 	cmd.Flags().StringSliceVar(&f.flagNamespaceConfigAnnotations, flagNamespaceConfigAnnotations, nil, "Namespace configuration annotations in form key=value.")
 	cmd.Flags().StringSliceVar(&f.flagNamespaceConfigLabels, flagNamespaceConfigLabels, nil, "Namespace configuration labels in form key=value.")
+
+	_ = cmd.Flags().MarkHidden(flagOrganization)
```
pipo02mix, line 60: why do we hide it?

pipo02mix (approved):

cmd/template/app/flag.go
ljakimczuk: At the moment the org-namespaced apps work only on the CAPO clusters, and we hide CAPO-related flags in the `template cluster`, so we decided to hide this one too

PR #644, opened 2022-01-26
Use v1beta1 types on Azure

fiunchinho:
This PR updates the CAPI manifest used to template Azure clusters from v1alpha4 to v1beta1, which is the latest CAPI version.

[We pass the `allocate-node-cidrs` to controller-manager](https://github.com/giantswarm/kyverno-policies/pull/98) , and that requires to set the pod cidrs on the `Cluster` CR. This PR also adds a default for that.

### DIFF
--- a/cmd/template/cluster/provider/templates/azure/azure_machine_template.yaml.tmpl
+++ b/cmd/template/cluster/provider/templates/azure/azure_machine_template.yaml.tmpl
@@ -1,4 +1,4 @@
-apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
 kind: AzureMachineTemplate
 metadata:
   labels:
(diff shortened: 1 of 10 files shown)


whites11 (approved):

cmd/template/cluster/provider/templates/azure/azure_machine_template.yaml.tmpl
```diff
@@ -11,6 +11,11 @@ metadata:
   namespace: {{ .Namespace }}
 spec:
   template:
+    metadata:
+      labels:
+        "release.giantswarm.io/version": "{{ .Version }}"
+        "giantswarm.io/organization": "{{ .Organization }}"
+        "cluster.x-k8s.io/watch-filter": "capi"
```
fiunchinho, line 18: This is required or the `AzureMachine` created won't be reconciled by our capi controllers using the watch filter label.

anvddriesch (approved):

PR #649, opened 2022-01-31
Add catalog-namespace flag to template app

rossf7:
Follow up to https://github.com/giantswarm/giantswarm/issues/19941

> If the Catalog CR is not in the default namespace you need to set the catalog namespace to the organization namespace where it is stored. This can be done using kubectl gs template app.

I somehow missed in testing that this flag didn't exist yet. Getting it sorted.


MarcelMue (approved):
Makes sense to me.

PR #655, opened 2022-02-02
Drop client wrappers

tfussell:
This is purely a refactor which shouldn't affect code execution at all. I'll verify that all commands work identically before and after.

Summary of changes:
- Use `k8smetadata` instead of `apiextensions/pkg/labels` and `apiextensions/pkg/annotations`.
- Use `k8sclient` directly instead of wrapper in `pkg/data/client`.
- Use `k8sclient/pkg/k8sclienttest` instead of fake in `pkg/data/client`.
- Centralize scheme creation and scheme builder definition to `pkg/scheme`.
- Drop `runtimeclient` alias from `sigs.k8s.io/controller-runtime/pkg/client` to standardize naming with other projects.
- Return structs instead of interfaces from constructors.
- Move service fakes into `_test.go` files to avoid cluttering non-test code.
- Copy `apiextensions/pkg/id` into `pkg/id` to reduce dependency on `apiextensions`.


cmd/get/capi/runner.go
```diff
@@ -223,7 +226,7 @@ var (
 		},
 		{
 			DisplayName:   "OpenStack Provider",
-			LabelSelector: "app.kubernetes.io/name=cluster-api-provider-openstack",
+			LabelSelector: "cluster.x-k8s.io/provider=infrastructure-openstack,control-plane=capo-controller-manager",
```
tfussell, line 229: We use clusterctl on OpenStack for now so these labels were wrong.

cmd/get/capi/runner.go
```diff
@@ -274,19 +276,22 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
...
-	for _, crd := range crds {
-		foundCrd, err := client.ExtClient().ApiextensionsV1().CustomResourceDefinitions().Get(
-			ctx,
-			crd.Name,
-			metav1.GetOptions{},
-		)
-		if err != nil && !errors.IsNotFound(err) {
-			return microerror.Mask(err)
-		}
+	crdList, err := k8sClients.ExtClient().ApiextensionsV1().CustomResourceDefinitions().List(ctx, metav1.ListOptions{})
```
tfussell, line 279: Speed up the command by listing all CRDs instead of getting them one by one.

tfussell:
I tested the following commands against `gollum`:
- `get apps`
- `get catalogs`
- `get clusters`
- `get nodepools`
- `get releases`
- `login`
- `selfupdate`
- `template app`
- `template catalog`
- `template cluster --provider openstack`
- `template networkpool`
- `update app`
- `validate apps`

I didn't test `update cluster` because I don't know how.

anvddriesch (approved):
I may have missed something in this long PR but it looks good to me. Please make sure that in case of replacing `get` with `list` that there is no case in which we list across all org-namespaces since that would fail for some users.
What was the problem with testing `update cluster`? I think if you create a test cluster in an older version it should work.

tfussell:
> I may have missed something in this long PR but it looks good to me. Please make sure that in case of replacing `get` with `list` that there is no case in which we list across all org-namespaces since that would fail for some users. What was the problem with testing `update cluster`? I think if you create a test cluster in an older version it should work.

Yes, sorry it's so long. I was just applying the same changes to all the places where we use the kubernetes client.

Good point about the list permission. I will check that and report back.

I've never used `update cluster` before. I think it adds an annotation to the Cluster CR only? I will see if I can exercise it today.

tfussell:
Following up on the previous discussion:
- `update cluster` still works.
- `get capi` lists CRDs which are not namespaced so anything that worked before with `get` should still work with `list` (as long as the `Role`/`ClusterRole` allows it.

PR #661, opened 2022-02-04
add support for self-contained kubeconfig in MC case

anvddriesch:
Towards https://github.com/giantswarm/giantswarm/issues/20669
This adds the ability to create a self-contained kubeconfig file for MC contexts.


cmd/login/auth.go
```diff
@@ -24,9 +24,9 @@ type authInfo struct {
 	refreshToken string
 }
 
-// storeCredentials stores the installation's CA certificate, and
+// storeMCCredentials stores the installation's CA certificate, and
```
anvddriesch, line 27: I allowed myself to change the names of these commands to make them less confusing

cmd/login/auth.go
```diff
@@ -175,8 +254,9 @@ func switchContext(ctx context.Context, k8sConfigAccess clientcmd.ConfigAccess,
 	} else if authType == kubeconfig.AuthTypeUnknown {
 		return microerror.Maskf(incorrectConfigurationError, "There is no authentication configuration for the '%s' context", newContextName)
 	}
-
-	config.CurrentContext = newContextName
+	if !keepContext {
```
anvddriesch, line 257: We only switch context here if that is desired.

cmd/login/clientcert.go
```diff
@@ -201,7 +201,9 @@ func storeCredential(k8sConfigAccess clientcmd.ConfigAccess, fs afero.Fs, client
 		config.Contexts[contextName] = context
 
 		// Select newly created context as current.
-		config.CurrentContext = contextName
+		if !keepContext {
```
anvddriesch, line 204: We also want to be aware of the keep-context functionality in WC case

cmd/login/flag.go
```diff
@@ -19,7 +19,7 @@ const (
 	flagWCOrganization      = "organization"
 	flagWCCertGroups        = "certificate-group"
 	flagWCCertTTL           = "certificate-ttl"
-	flagWCSelfContained     = "self-contained"
+	flagSelfContained       = "self-contained"
```
anvddriesch, line 22: We rename this flag because it's being opened to MC cases

cmd/login/runner.go
```diff
@@ -181,12 +181,22 @@ func (r *runner) loginWithKubeContextName(ctx context.Context, contextName strin
 // one with the name derived from the installation code name.
 func (r *runner) loginWithCodeName(ctx context.Context, codeName string) error {
 	var contextAlreadySelected bool
+	var newLoginRequired bool
+	var selfContained bool
 
 	contextName := kubeconfig.GenerateKubeContextName(codeName)
-	err := switchContext(ctx, r.k8sConfigAccess, contextName)
+	err := switchContext(ctx, r.k8sConfigAccess, contextName, r.flag.KeepContext || (len(r.flag.SelfContained) > 0))
```
anvddriesch, line 188: We also keep the context in case we create a self-contained kubeconfig

cmd/login/runner.go
```diff
@@ -181,12 +181,22 @@ func (r *runner) loginWithKubeContextName(ctx context.Context, contextName strin
...
 	if IsContextAlreadySelected(err) {
 		contextAlreadySelected = true
 	} else if IsNewLoginRequired(err) {
+		newLoginRequired = true
+	} else if err != nil {
+		return microerror.Mask(err)
+	}
+
+	selfContained = len(r.flag.SelfContained) > 0 && !(len(r.flag.WCName) > 0)
```
anvddriesch, line 197: We only create the self-contained kubeconfig here if only the MC is requested. (Otherwise it would be overwritten later when WC client cert is created)

cmd/login/runner.go
```diff
@@ -181,12 +181,22 @@ func (r *runner) loginWithKubeContextName(ctx context.Context, contextName strin
...
+
+	if newLoginRequired || selfContained {
```
anvddriesch, line 199: We create the self-contained MC kubeconfig from scratch

cmd/login/runner.go
```diff
@@ -205,13 +215,11 @@ func (r *runner) loginWithCodeName(ctx context.Context, codeName string) error {
...
 		return nil
-	} else if err != nil {
-		return microerror.Mask(err)
 	}
 
 	if contextAlreadySelected {
 		fmt.Fprintf(r.stdout, "Context '%s' is already selected.\n", contextName)
-	} else {
+	} else if !(r.flag.KeepContext || selfContained) {
 		fmt.Fprintf(r.stdout, "Switched to context '%s'.\n", contextName)
```
anvddriesch, line 223: TODO: go through all of these info statements and add the needed information that a self-contained kubeconfig was created and also make sure there are no inaccurate "switch context" statements anymore.

cmd/login/runner.go
```diff
@@ -181,12 +181,22 @@ func (r *runner) loginWithKubeContextName(ctx context.Context, contextName strin
...
+	var newLoginRequired bool
+	var selfContained bool
 
 	contextName := kubeconfig.GenerateKubeContextName(codeName)
-	err := switchContext(ctx, r.k8sConfigAccess, contextName)
+	err := switchContext(ctx, r.k8sConfigAccess, contextName, r.flag.KeepContext || (len(r.flag.SelfContained) > 0))
 	if IsContextAlreadySelected(err) {
 		contextAlreadySelected = true
 	} else if IsNewLoginRequired(err) {
+		newLoginRequired = true
```
kuosandys, line 192: I noticed while testing that this condition isn't triggered, so when I try to login using a code name with the self-contained flag (`kubectl gs login <codename> --self-contained`), I get a "Token renewal failed" error because there's no refresh token in the empty file.

I think `switchContext()` isn't returning a `newLoginRequiredError` though it does return `tokenRenewalFailedError` when renewing the token fails?

anvddriesch: I tried this and it happened to me, too. Do you think it would be safe to continue with the new login despite the error in case we are requesting a self contained kubeconfig?

kuosandys: I think that makes sense 👍

anvddriesch: Let me know if you still see the issue.

kuosandys: Not anymore - looks good! Thanks for checking this out 😄

kuosandys (approved):
👍

PR #666, opened 2022-02-07
add config for options

anvddriesch:
This is a sub PR to https://github.com/giantswarm/kubectl-gs/pull/661 to make things a bit more visible
We add a little config that contains options to determine how to switch back and forth between contexts.
This should also fix the bug you found @kuosandys 


cmd/login/runner.go
```diff
@@ -31,11 +31,21 @@ type runner struct {
 	fs     afero.Fs
 
 	k8sConfigAccess clientcmd.ConfigAccess
+	loginOptions    LoginOptions
 
 	stdout io.Writer
 	stderr io.Writer
 }
 
+type LoginOptions struct {
```
anvddriesch, line 40: This could also be extended with other options. For now I used the ones that were important to the task at hand.

kuosandys (approved):
This makes the different options easier to understand! Thank you ❤️

PR #667, opened 2022-02-07
Add support for openstack to get cluster command

tfussell:
_Description goes here_

---

### DIFF
--- a/pkg/commonconfig/commonconfig.go
+++ b/pkg/commonconfig/commonconfig.go
@@ -47,7 +47,7 @@ func (cc *CommonConfig) GetProvider() (string, error) {
 		provider = key.ProviderAzure
 
 	default:
-		provider = key.ProviderKVM
+		provider = key.ProviderOpenStack
 	}
 
 	return provider, nil
--- a/pkg/data/domain/cluster/openstack.go
+++ b/pkg/data/domain/cluster/openstack.go
@@ -1,0 +1,80 @@
+package cluster
+
+import (
+	"context"
+
+	"github.com/giantswarm/microerror"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	capiv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
+	runtimeClient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func (s *Service) getAllOpenStack(ctx context.Context, namespace string) (Resource, error) {
+	var err error
+
+	inNamespace := runtimeClient.InNamespace(namespace)
+
+	clusters := &capiv1alpha3.ClusterList{}
+	{
+		err = s.client.List(ctx, clusters, inNamespace)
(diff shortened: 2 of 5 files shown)


marians:
⚠️ **Headsup! Branch `master` has been renamed to `main`**

```nohighlights
git branch -m master main
git fetch origin
git remote set-head origin -a
```

tfussell:
I'm not adding this to the docs currently until we have the output structure nailed down for helm-based clusters.

cmd/get/clusters/provider/common.go
```diff
@@ -11,3 +15,32 @@ const (
 func formatCondition(condition string) string {
 	return strings.ToUpper(condition)
 }
+
+func getLatestCondition(conditions []capi.Condition) string {
```
pipo02mix, line 19: nice refactoring

pkg/commonconfig/commonconfig.go
```diff
@@ -47,7 +47,7 @@ func (cc *CommonConfig) GetProvider() (string, error) {
 		provider = key.ProviderAzure
 
 	default:
-		provider = key.ProviderKVM
+		provider = key.ProviderOpenStack
```
pipo02mix, line 50: I am not sure about this one. Should we dont leave KVM as default and a regexp for OpenStack?

pipo02mix: hmm I see we don't have a way to know it is OpenStack from the host, right? but how do we deal with GCP or another provider? it seems this is not robust enough

pkg/data/domain/cluster/openstack.go
```diff
@@ -0,0 +1,153 @@
...
+		if apierrors.IsForbidden(err) {
+			return nil, microerror.Mask(insufficientPermissionsError)
+		} else if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		capiCluster.TypeMeta = meta.TypeMeta{
+			APIVersion: "cluster.x-k8s.io/v1alpha3",
+			Kind:       "Cluster",
+		}
```
pipo02mix, line 106: why do we need to set up the APIversion? 🤔

pipo02mix (commented):
some questions

pkg/commonconfig/commonconfig.go
tfussell: Exactly. `kubectl-gs` doesn't support KVM so this felt fine to me for now. We can deal with GCP when we need it. The solution will be coming up with a way to determine the provider of an MC just using a kubeconfig. That could be using athena or maybe a metadata CM that lives in a known place on the MC.

pkg/data/domain/cluster/openstack.go
tfussell: I think this was added so that `yaml.Marshal` shows the API version correctly when objects are generated client-side (Kubernetes API injects this automatically when you retrieve an object, but if you create it in code it won't be set). It felt unnecessary as these are retrieved from the API, but I didn't want to make too many changes here so I'll try to remove in a separate PR.

pipo02mix (approved):

PR #651, opened 2022-02-01
Build openstack app values from flags

tfussell:
This PR is intended to improve the UX of OpenStack cluster creation. It:
- Drops the `--cluster-app-values` and `--defaults-apps-app-values` flags in favor of building the ConfigMap values entirely dynamically.
- Fetches the latest release of cluster-openstack and default-apps-openstack from the management cluster `AppCatalogEntry` if not specified.
- Adds basic validation for required OpenStack flags.
- Drops `--cluster-topology` flag and non-app-based OpenStack cluster templates.
- Renames `--cluster-app-version` to `--cluster-version`.
- Renames `--cluster-app-catalog` to `--cluster-catalog`.
- Renames `--default-apps-app-version` to `--default-apps-version`.
- Renames `--default-apps-app-catalog` to `--default-apps-catalog`.
- Adds new flags for all important values in https://github.com/giantswarm/cluster-openstack/blob/main/helm/cluster-openstack/values.yaml.
- Drops duplicated provider flags struct in favor of shared structs for passing around values in the provider package.
- Break up `WriteOpenStackTemplateAppCR` into a function for cluster-openstack and a function for default-apps-openstack.

This is not in the changelog as OpenStack is still in alpha and all relevant flags were already hidden.

Example command `kubectl gs template cluster --provider openstack --organization multi-project --enable-oidc --failure-domain test-1 --cloud openstack --cloud-config cloud-config-giantswarm-2 --bastion-image-uuid be59b881-a5a8-4785-85d1-f0c8743bc20d --node-image-uuid 7cd94dd0-2815-427c-ac25-9a1708645dc5 --external-network-id bb5da432-3749-45f5-a1ab-a10eaa5a8a13 --node-cidr 10.6.0.0/24 --bastion-machine-flavor n1.tiny --control-plane-machine-flavor n1.small --worker-machine-flavor n1.small`

Example output
```yaml
---
apiVersion: v1
data:
  values: |
    bastion:
      flavor: n1.tiny
      image: ""
      rootVolume:
        diskSize: 10
        sourceUUID: be59b881-a5a8-4785-85d1-f0c8743bc20d
    cloudConfig: cloud-config-giantswarm-2
    cloudName: openstack
    controlPlane:
      diskSize: 10
      machineFlavor: n1.small
      replicas: 1
    externalNetworkID: bb5da432-3749-45f5-a1ab-a10eaa5a8a13
    nodeCIDR: 10.6.0.0/24
    nodeClasses:
    - diskSize: 10
[...]

PR #668, opened 2022-02-07
Fix cluster catalog flag

tfussell:
Fixing copy and paste error from https://github.com/giantswarm/kubectl-gs/pull/651


tfussell:
Going with a ping.

marians:
Changelog entry would be nice

PR #676, opened 2022-02-11
capa-use-new-release-repos

calvix:


cmd/template/cluster/provider/capo.go
```diff
@@ -202,24 +199,3 @@ func templateDefaultAppsOpenstack(ctx context.Context, k8sClient k8sclient.Inter
...
-
-	if err != nil {
-		return "", microerror.Mask(err)
-	} else if len(catalogEntryList.Items) != 1 {
-		message := fmt.Sprintf("version not specified for %s and latest release couldn't be determined in %s catalog", app, catalog)
-		return "", microerror.Maskf(invalidFlagError, message)
-	}
-
-	return catalogEntryList.Items[0].Spec.Version, nil
-}
```
calvix, line 225: moved to common file as its now used by both CAPO and CAPA

marians:
⚠️ **Headsup! Branch `master` has been renamed to `main`**

```nohighlights
git branch -m master main
git fetch origin
git remote set-head origin -a
```

pkg/template/app/app.go
```diff
@@ -57,6 +57,7 @@ func NewAppCR(config Config) ([]byte, error) {
 	if config.InCluster {
 		crNamespace = config.Namespace
 		appLabels[label.AppOperatorVersion] = "0.0.0"
+		appLabels[label.ManagedBy] = "flux"
```
tfussell, line 60: I think we only want this for Flux-managed clusters, right @MarcelMue? Maybe we should add a flag like "--cluster-app-extra-labels` so arbitrary labels could be passed in and it would be up to the user to set this.

cmd/template/cluster/provider/templates/capa/types.go
```diff
@@ -0,0 +1,53 @@
+package capa
+
+type ClusterConfig struct {
```
tfussell, line 3: I feel like we could structure this better so the common values are in one package and the provider-specific values are in the provider packages. Fine as a start though.

cmd/template/cluster/provider/common.go
```diff
@@ -3,29 +3,54 @@ package provider
...
+	BastionInstanceType      string
+	BastionReplicas          int
+	ControlPlaneInstanceType string
+	SSHSSOPublicKey          string
+}
+
+type AWSMachinePoolConfig struct {
+	Name             string
+	Min              int
+	Max              int
```
tfussell, line 49:
```suggestion
	MinSize          int
	MaxSize          int
```

cmd/template/cluster/provider/templates/aws/template.go
```diff
@@ -4,57 +4,20 @@ import (
...
-		{Name: "aws_machine_template.yaml.tmpl", Data: awsMachineTemplate},
-		{Name: "aws_cluster_role_identity.yaml.tmpl", Data: awsClusterRoleIdentity},
-		{Name: "bastion_secret.yaml.tmpl", Data: bastionSecret},
-		{Name: "bastion_machine_deployment.yaml.tmpl", Data: bastionMachineDeployment},
-		{Name: "bastion_aws_machine_template.yaml.tmpl", Data: bastionAWSMachineTemplate},
-	}
-}
-
 // GetEKSTemplate merges .tmpl files for an EKS cluster.
 func GetEKSTemplates() []Template {
```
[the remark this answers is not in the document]
calvix: As there is almost no shared CRs for EKS I think we will make a separate repo for EKS

pkg/template/app/app.go
calvix: I think I added this as a recommendation from some slack thread, not sure if it's necessary or not.

cmd/template/cluster/flag.go
```diff
@@ -24,6 +24,21 @@ const (
...
+	flagBastionInstanceType            = "bastion-instance-type"
+	flagBastionReplicas                = "bastion-replicas"
+	flagControlPlaneInstanceType       = "control-plane-instance-type"
+	flagAWSMachinePoolMinSize          = "machine-pool-min-size"
+	flagAWSMachinePoolMaxSize          = "machine-pool-max-size"
+	flagAWSMachinePoolName             = "machine-pool-name"
+	flagAWSMachinePoolAZs              = "machine-pool-azs"
+	flagAWSMachinePoolInstanceType     = "machine-pool-instance-type"
+	flagAWSMachinePoolRootVolumeSizeGB = "machine-pool-root-volume-size-gb"
+	flagAWSMachinePoolCustomNodeLabels = "machine-pool-custom-node-labels"
```
fiunchinho, line 40: It sounds like some of these can be shared across providers, like `bastion-instance-type` or `bastion-replicas` or `control-plane-instance-type` or even `region` (if I agree on using that term for all providers). Any reason not to share them?

calvix: the naming terminology is not consistent across providers so I am not sure how good UX is to mix them, on was we have instance type, on OpenStack there is flavor, Not sure how would someone who never worked with  Openstack created if you need to select bastion flavor for AWS

calvix: the whole flag is a mess in my opinion and I don't know a good  approach how to get it straight and have a good UX for it

calvix: I think sorting all flags across all provider should not be scope fo this PR, but rather some big story for KaaS sync

cmd/template/cluster/flag.go
```diff
@@ -243,6 +290,22 @@ func (f *flag) Validate() error {
...
+			}
+			if isCapiVersion {
+				if f.AWS.Region == "" {
+					return microerror.Maskf(invalidFlagError, "--%s is required", flagAWSRegion)
+				}
+				if f.ControlPlaneAZ == nil {
+					return microerror.Maskf(invalidFlagError, "--%s is required", flagControlPlaneAZ)
+				}
+				if f.AWS.MachinePool.AZs == nil {
+					return microerror.Maskf(invalidFlagError, "--%s is required", flagControlPlaneAZ)
```
AverageMarcus, line 305: Is `flagControlPlaneAZ` a typo here?

fiunchinho (approved):

cmd/template/cluster/flag.go
calvix: yep, copy-past typo
nice catch

calvix:
I created a note about the flags in the KaaS sync so we can talk about the next steps

AverageMarcus (approved):

PR #688, opened 2022-02-24
Fix CAPZ templates

fiunchinho:
ClusterAPI creates CRs from other CRs that act as templates. The `KubeadmControlPlane` is used as template to create the `KubeadmConfig`. When deploying the control plane nodes, CAPI will check if the control plane `KubeadmConfig` matches the template found at the `KubeadmControlPlane` CR. If they don't match, it means an upgrade needs to be done and nodes are rolled.
In our current setup kyverno will default some fields in the `KubeadmConfig` CR. This means that the resulting `KubeadmConfig` CR won't be exactly the same as its source template in the `KubeadmControlPlane` CR, making CAPI think that nodes need to be roll forever.

In this PR I'm adding those fields to the template in the `KubeadmControlPlane` CR.


cmd/template/cluster/provider/templates/azure/azure_machine_template.yaml.tmpl
```diff
@@ -21,7 +21,7 @@ spec:
       - diskSizeGB: 256
         lun: 0
         nameSuffix: etcddisk
-      identity: SystemAssigned
+      identity: None
```
whites11, line 24: ?

whites11 (approved):

cmd/template/cluster/provider/templates/azure/azure_machine_template.yaml.tmpl
fiunchinho: my reasoning is that using SystemAssigned identity will fail if a new machine is added for whatever reason. The new machine will never work. My upstream PR got merged but it's still not released. So I thought that using None would help here. Using None means using the env vars that are passed to the capz controllers, I figured that was good enough for our alpha release

PR #693, opened 2022-03-03
Add enable-long-names feature flag

tfussell:
Following discussion in KaaS Sync today, I thought we could start by relaxing the name restrictions in `kubectl-gs` using a hidden feature flag. Anyone who would like to test can add `--enable-long-names` to any command that validates or generates a name. There should be no impact on other users.

---


pkg/id/id.go
```diff
@@ -1,37 +0,0 @@
-package id
```
tfussell, line 1: I added this package not realizing the same function already existing in `key`.

internal/key/key.go
```diff
@@ -23,43 +22,29 @@ import (
...
+	// NameLengthLong represents the number of characters used to create a cluster ID when --enable-long-names feature flag is used.
+	NameLengthLong = 10
+	// NameLengthShort represents the number of characters used to create a cluster ID.
+	NameLengthShort = 5
 
 	organizationNamespaceFormat = "org-%s"
 )
 
 const (
-	AWSBastionInstanceType = "t3.small"
```
tfussell, line 36: Unrelated. Removing unused constants and functions in this package.

internal/key/key.go
```diff
@@ -68,63 +53,46 @@ func NodeSSHDConfigEncoded() string {
...
+		for i := range characters {
+			characters[i] = letterRunes[rand.Intn(len(letterRunes))] //nolint:gosec
 		}
 
-		id := string(b)
+		generatedName := string(characters)
 
-		if _, err := strconv.Atoi(id); err == nil {
-			// ID is made up of numbers only, which we want to avoid.
+		if valid, err := ValidateName(generatedName, enableLongNames); err != nil {
```
tfussell, line 82: Reusing the same validation function as we use to validate a user-provided name instead of having a separate regex.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
+- Add `--enable-long-names` feature flag to allow resource names longer than 5 characters. Only for internal testing.
```
marians, line 12: Please add to which command the flag is being added

pipo02mix (approved):
LGTM

CHANGELOG.md
tfussell: Done.

tfussell:
> Thanks for the refactoring!
> 
> Could you please explain which commands are affected and why? From the KaaS sync context I would have expected impact on commands related to clusters and node pools only, but not `template catalog`.

@marians After discussing with Honeybadger, I have removed the changes to the `template catalog` command.

marians (approved):

PR #698, opened 2022-03-07
Adding description column

ljakimczuk:
Add `DESCRIPTION` column to the `get catalog` command.

Towards: https://github.com/giantswarm/giantswarm/issues/21129


cmd/get/catalogs/printer.go
```diff
@@ -61,6 +61,14 @@ func (r *runner) printNoResourcesOutput() {
 	fmt.Fprintf(r.stdout, "  kubectl gs template catalog --help\n")
 }
 
+func getAppCatalogEntryDescription(description string) string {
+	if len(description) > 80 {
```
ljakimczuk, line 65: should this be fixed or rather configurable via flag?

anvddriesch: Do you mean the character limit?
If you think there is a use case for it, feel free to add a flag and set the default to 80.

anvddriesch (approved):

marians (approved):

marians:
⚠️ *Headsup! Branch `master` has been renamed to `main`*

```nohighlights
git branch -m master main
git fetch origin
git remote set-head origin -a
```

cmd/get/catalogs/printer.go
rossf7: I don't think its a typical use case but I could see a flag being useful for truncating or extending the output. 

@ljakimczuk I agree with @anvddriesch feel free to add if you think its needed.

rossf7 (approved):

cmd/get/catalogs/printer.go
ljakimczuk: @rossf7 @anvddriesch what do you think about the current approach? Flag is persistent so it can be re-used with other get sub-commands. I don't like passing the argument down the stream to the function that need it though, it could be set as a global variable, but I am not sure what's the right style/approach 😅

rossf7: @ljakimczuk I would add it to flag.go like I described here.
https://github.com/giantswarm/kubectl-gs/pull/698/files#r821565602

cmd/get/catalogs/runner.go
```diff
@@ -107,7 +107,15 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		}
 	}
 
-	err = r.printOutput(catalogResource)
+	var maxColWidth uint
+	{
+		maxColWidth, err = cmd.InheritedFlags().GetUint("max-col-width")
+		if err != nil {
+			return microerror.Mask(err)
+		}
```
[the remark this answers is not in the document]
ljakimczuk: @rossf7 is just that I wanted it to be sort of global to all the get sub-commands without definining it in each of them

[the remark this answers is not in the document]
rossf7: Do we need it in the other get sub-commands?

You could use the same approach we use to add the upstream print flags. Otherwise not sure how to do this.

https://github.com/giantswarm/kubectl-gs/blob/5e7577ad47e575c7557208800a3b6d81da895c60/cmd/get/catalogs/flag.go#L25-L28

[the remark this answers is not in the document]
ljakimczuk: > Do we need it in the other get sub-commands?
Well, not sure, sounds like it should be common to all the command that prints something, but anyway, I don't see any problem moving it downstream to the sub-command that currently uses it. Anyway it just a cosmetic change and we can figure out how to make it common when we will actually need it.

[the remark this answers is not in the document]
rossf7: Cool this looks good and if we need this in other commands we can adjust later.

rossf7 (approved):

PR #708, opened 2022-03-10
Fetch release components to set CR labels when templating

fiunchinho:
Towards https://github.com/giantswarm/giantswarm/issues/21251

Since the very first release of `kubectl-gs` we have code that _tries_ to add the labels for the GiantSwarm components (to know which resources to reconcile) to the cluster templates. This has been always broken, but we haven't noticed until now because we have our own admission controllers that also default these labels. So the templates generated by kubectl-gs were wrong, but the admission webhooks set the right values on these labels.

We should generate the right templates from `kubectl-gs`.

### DIFF
--- a/cmd/template/cluster/provider/aws.go
+++ b/cmd/template/cluster/provider/aws.go
@@ -35,7 +35,7 @@ func WriteAWSTemplate(ctx context.Context, client k8sclient.Interface, out io.Wr
 			}
 		}
 	} else {
-		err = WriteGSAWSTemplate(out, config)
+		err = WriteGSAWSTemplate(ctx, client, out, config)
 		if err != nil {
 			return microerror.Mask(err)
 		}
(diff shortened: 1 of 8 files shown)


cmd/template/cluster/provider/aws.go
```diff
@@ -59,6 +59,11 @@ func WriteGSAWSTemplate(out io.Writer, config ClusterConfig) error {
 		Labels:         config.Labels,
 	}
 
+	crsConfig.ReleaseComponents, err = key.GetReleaseComponents(ctx, client.CtrlClient(), config.ReleaseVersion)
```
fiunchinho, line 62: The `ReleaseComponents` field was being referenced from the templates, but it was never initialized. So just initializing it with the right values is enough to fix this.

internal/key/key.go
```diff
@@ -174,6 +175,21 @@ func SSHSSOPublicKey(ctx context.Context, client runtimeclient.Client) (string,
 	return sshSSOPublicKey, nil
 }
 
+func GetReleaseComponents(ctx context.Context, client runtimeclient.Client, releaseName string) (map[string]string, error) {
```
fiunchinho, line 178: We are using this function from two different commands (`template cluster` and `template nodepool` and for `aws` and `azure`) so I decided to put it on the `key` package.

AverageMarcus (approved):

PR #710, opened 2022-03-11
Add visibility flag to template catalog

rossf7:
Follow up to https://github.com/giantswarm/giantswarm/issues/19941

## Reason

As a user of app platform I'd like to be able to view custom catalogs in the web UI.

## Context

Only catalog CRs with the label `application.giantswarm.io/catalog-visibility: public` are shown. Any other value hides the catalog unless the user is an admin.

This adds a `--visibility` flag that defaults to public. That makes the most sense to me but happy to change the default if others disagree.

Docs: https://github.com/giantswarm/docs/pull/1361

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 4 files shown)


MarcelMue (approved):
Seems pretty straightforward to me.

PR #718, opened 2022-03-21
Add missing fix to changelog

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/21161 and discovered while testing https://github.com/giantswarm/test-infra/pull/164

A bug in `kubectl-gs` v2.3.1 was fixed with https://github.com/giantswarm/kubectl-gs/pull/676 and released as v2.4.0, but it wasn't mentioned in the changelog. I found the change using `git diff` and thought it was worth mentioning in the changelog even though it's already released.


tfussell:
Going with a ping as this is purely informational.

PR #723, opened 2022-03-24
support clientcert creation for wc on openstack

anvddriesch:
Towards https://github.com/giantswarm/roadmap/issues/884

This adds clientCert creation for clusters in the openstack provider.
Also I did some refactoring to have a bit less spaghetti but it resulted in a messy PR, sorry about that. I will add comments to create more clarity.


cmd/login/clientcert.go
```diff
@@ -48,6 +55,23 @@ type clientCertConfig struct {
 	certOperatorVersion string
 }
 
+type serviceSet struct {
```
anvddriesch, line 58: Adding some structs to reduce number of inputs.

cmd/login/clientcert.go
```diff
@@ -139,22 +157,100 @@ func fetchCredential(ctx context.Context, provider string, clientCertService cli
 	return secret, nil
 }
 
+func generateCredential(ctx context.Context, ca *corev1.Secret, config clientCertConfig) (*corev1.Secret, error) {
```
anvddriesch, line 160: This is the function we use to create the client certificate for openstack clusters. We sign it using the CA secret for the WC.

cmd/login/clientcert.go
```diff
@@ -139,22 +157,100 @@ func fetchCredential(ctx context.Context, provider string, clientCertService cli
...
+	})
+}
+func getPrivKey(keyPEM []byte) (*rsa.PrivateKey, error) {
+	block, _ := pem.Decode(keyPEM)
+	return x509.ParsePKCS1PrivateKey(block.Bytes)
+}
+
 // storeWCCredentials saves the created client certificate credentials into the kubectl config.
-func storeWCCredentials(k8sConfigAccess clientcmd.ConfigAccess, fs afero.Fs, clientCert *clientcert.ClientCert, credential *corev1.Secret, clusterBasePath string, loginOptions LoginOptions) (string, bool, error) {
+func storeWCCredentials(k8sConfigAccess clientcmd.ConfigAccess, fs afero.Fs, c credentialConfig) (string, bool, error) {
```
anvddriesch, line 244: Added a config here to simplify a bit

cmd/login/clientcert.go
```diff
@@ -217,22 +313,22 @@ func storeWCCredentials(k8sConfigAccess clientcmd.ConfigAccess, fs afero.Fs, cli
 }
 
 // printWCCredentials saves the created client certificate credentials into a separate kubectl config file.
-func printWCCredentials(k8sConfigAccess clientcmd.ConfigAccess, fs afero.Fs, filePath string, clientCert *clientcert.ClientCert, credential *corev1.Secret, clusterBasePath string, loginOptions LoginOptions) (string, bool, error) {
+func printWCCredentials(k8sConfigAccess clientcmd.ConfigAccess, fs afero.Fs, c credentialConfig) (string, bool, error) {
```
anvddriesch, line 316: Same as above

cmd/login/wc.go
```diff
@@ -19,48 +21,17 @@ import (
 	"github.com/giantswarm/kubectl-gs/pkg/kubeconfig"
 )
 
-func (r *runner) handleWCLogin(ctx context.Context) error {
-	// At the moment, the only available login option for WC is client cert
-	return r.handleWCClientCert(ctx)
-}
-
-func (r *runner) handleWCClientCert(ctx context.Context) error {
+func (r *runner) getServiceSet(client k8sclient.Interface) (serviceSet, error) {
```
anvddriesch, line 24: Moving the service creation to its own function

cmd/login/wc.go
```diff
@@ -90,52 +61,134 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
...
 
+	return serviceSet{
+		clientCertService:   clientCertService,
+		organizationService: organizationService,
+		clusterService:      clusterService,
+		releaseService:      releaseService,
+	}, nil
+}
+
+func (r *runner) getCluster(ctx context.Context, services serviceSet, provider string) (*cluster.Cluster, error) {
```
anvddriesch, line 76: moving the logic for getting the cluster into its own function

cmd/login/wc.go
```diff
@@ -90,52 +61,134 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
...
+	}
+
+	c, err = findCluster(ctx, services.clusterService, services.organizationService, provider, r.flag.WCName, namespaces...)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	return c, nil
+}
+
+func (r *runner) getCertOperatorVersion(c *cluster.Cluster, provider string, services serviceSet, ctx context.Context) (string, error) {
```
anvddriesch, line 106: Moving logic to get cert operator version into its own function. In openstack this is not needed and therefore empty.

cmd/login/wc.go
```diff
@@ -147,65 +200,90 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
 		certOperatorVersion: certOperatorVersion,
 	}
 
-	clientCertResource, err := createCert(ctx, clientCertService, certConfig)
+	clientCertResource, secret, err := r.getCredentials(ctx, services.clientCertService, certConfig)
```
anvddriesch, line 203: Moving handling of certificate and secret to one function

cmd/login/wc.go
```diff
@@ -147,65 +200,90 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
...
-		return microerror.Mask(err)
+		return "", false, microerror.Mask(err)
 	}
 
-	// Store client certificate credential either into the current kubeconfig or a self-contained file if a path is given.
-	var contextExists bool
-	var contextName string
-	if r.loginOptions.selfContainedWC {
-		contextName, contextExists, err = printWCCredentials(r.k8sConfigAccess, r.fs, r.flag.SelfContained, clientCertResource, secret, clusterBasePath, r.loginOptions)
+	if clientCertResource != nil {
```
anvddriesch, line 223: We only clean up if a clientCertResource has been created (non openstack case)

cmd/login/wc.go
```diff
@@ -147,65 +200,90 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
...
+	fmt.Fprint(r.stdout, color.GreenString("\nCreated client certificate for workload cluster '%s'.\n", r.flag.WCName))
+	return contextName, contextExists, nil
+}
+
+func (r *runner) getCredentials(ctx context.Context, clientCertService clientcert.Interface, config clientCertConfig) (*clientcert.ClientCert, *v1.Secret, error) {
+	var secret *v1.Secret
+	var clientCert *clientcert.ClientCert
+	var err error
+
+	if config.provider == key.ProviderOpenStack {
```
anvddriesch, line 240: in the openstack case we we fetch the ca secret and then generate the credentials

cmd/login/wc.go
```diff
@@ -147,65 +200,90 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
...
+		secret, err = generateCredential(ctx, ca, config)
+		if err != nil {
+			return nil, nil, microerror.Mask(err)
+		}
+		return nil, secret, nil
 	}
 
-	// Cleaning up leftover resources.
-	err = cleanUpClientCertResources(ctx, clientCertService, clientCertResource)
+	clientCert, err = createCert(ctx, clientCertService, config)
```
anvddriesch, line 253: Otherwise we do the same as before (applying a certConfig and fetching the credentials that have been created by certOperator)

cmd/login/clientcert.go
```diff
@@ -446,11 +542,11 @@ func getCertOperatorVersion(ctx context.Context, releaseService release.Interfac
 }
 
 func validateProvider(provider string) error {
-	if provider == key.ProviderAWS || provider == key.ProviderAzure {
-		return nil
+	if provider != key.ProviderAWS && provider != key.ProviderAzure && provider != key.ProviderOpenStack {
```
tfussell, line 545: Maybe it makes sense to check for providers that aren't supported instead of those that are--I think it would just be KVM.

anvddriesch: Yeah I did it that way at first.
However, new providers are now being added more frequently (like google the other day) and it might be better to consciously enable them IMO

cmd/login/wc.go
tfussell, line 240: I would prefer a capability based check here instead even if it's the same logic. Maybe a function like `isPureCAPI` or `usesMCPKI` which returns true for `openstack` and for `aws` and `azure` when release version is greater than or equal to `v20.0.0`.

cmd/login/wc.go
```diff
@@ -147,65 +200,90 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
...
+		ca, err := clientCertService.GetCredential(ctx, config.clusterNamespace, config.clusterName+"-ca")
 		if err != nil {
-			return microerror.Mask(err)
+			return nil, nil, microerror.Mask(err)
 		}
+		secret, err = generateCredential(ctx, ca, config)
+		if err != nil {
+			return nil, nil, microerror.Mask(err)
+		}
+		return nil, secret, nil
```
tfussell, line 250: Why don't you generate the client cert here as you do below for other providers?

cmd/login/wc.go
```diff
@@ -147,65 +200,90 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
...
 
-func getClusterBasePath(k8sConfigAccess clientcmd.ConfigAccess) (string, error) {
+func getClusterBasePath(k8sConfigAccess clientcmd.ConfigAccess, provider string) (string, error) {
 	config, err := k8sConfigAccess.GetStartingConfig()
 	if err != nil {
 		return "", microerror.Mask(err)
 	}
 
 	clusterServer, _ := kubeconfig.GetClusterServer(config, config.CurrentContext)
+	clusterName := kubeconfig.GetCodeNameFromKubeContext(config.CurrentContext)
```
tfussell, line 286: This function should maybe return an error if the context doesn't match the expected format. Also the kubeconfig from LastPass has a context that looks like `gamma-admin@gamma` which wouldn't work here (it seems to only accept the SSO context like `gs-gamma`).

cmd/login/clientcert.go
tfussell: All new providers will be pure CAPI so I feel the opposite--it's an extra chore to manually allow them versus allowing them by default.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 ## [Unreleased]
 
 - Add templating for CAPG clusters.
+- Support `openstack` provider in `login` command
```
marians, line 11:
```suggestion
- In `kubectl gs login`, add support for workload clusters on OpenStack.
```

cmd/login/clientcert.go
anvddriesch: Yeah, this makes sense if we replace the checks for openstack with a check for "pure capi" as you suggested below. :)

cmd/login/wc.go
anvddriesch: Thank you for that suggestion, I will add this. :)

cmd/login/wc.go
anvddriesch: The function that is used for other providers does not directly create the client certificate. It creates a `certConfig` which is applied to the MC. `cert-operator` reconciles it and then creates the certificate which is fetched. Later we need to clean up the `certConfig` that was created. (that is why we return it)
Since `cert-operator` is not part of CAPO, this does not work out of the box. We could do something similar, but I don't think this approach of creating something on the MC aligns with the gitops principle.
Let me know if there are better options, though.

pkg/data/domain/cluster/openstack.go
```diff
@@ -97,7 +97,7 @@ func (s *Service) getByNameOpenStack(ctx context.Context, name, namespace string
 		if apierrors.IsForbidden(err) {
 			return nil, microerror.Mask(insufficientPermissionsError)
 		} else if err != nil {
-			return nil, microerror.Mask(err)
+			return nil, microerror.Mask(notFoundError)
```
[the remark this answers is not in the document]
anvddriesch: Sure, I'll add that. :)

cmd/login/wc.go
anvddriesch: Good point! :)
I renamed the gamma context and didn't catch that one.

cmd/login/wc.go
```diff
@@ -147,59 +204,84 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
...
+	return contextName, contextExists, nil
+}
+
+func (r *runner) getCredentials(ctx context.Context, clientCertService clientcert.Interface, config clientCertConfig) (*clientcert.ClientCert, *v1.Secret, error) {
+	var secret *v1.Secret
+	var clientCert *clientcert.ClientCert
+	var err error
+
+	// If cert-operator is not running (as in CAPI) we attempt to use the MC PKI to create a certificate
+	if config.certOperatorVersion == "" {
```
anvddriesch, line 245: @tfussell I looked into implementing a `PureCAPI` check here but it would add a lot more complexity to this PR. 
For now, trying to use the MCPKI method in case `cert-operator` is not present makes more sense IMO since that is the reason the old implementation does not work for CAPI releases. WDYT?

cmd/login/wc.go
```diff
@@ -90,52 +61,138 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
...
+
+	c, err = findCluster(ctx, services.clusterService, services.organizationService, provider, r.flag.WCName, namespaces...)
+	if err != nil {
+		return nil, microerror.Mask(err)
+	}
+	return c, nil
+}
+
+func (r *runner) getCertOperatorVersion(c *cluster.Cluster, provider string, services serviceSet, ctx context.Context) (string, error) {
 	releaseVersion, err := getClusterReleaseVersion(c, provider, r.flag.WCInsecureNamespace)
```
tfussell, line 107: This line of code will still be executed for pure CAPI clusters (e.g. OpenStack), right? If so, OpenStack clusters don't have a release label so an error will be returned and this function will fail (i.e. won't get to line 115).

cmd/login/wc.go
```diff
@@ -90,52 +61,138 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
...
+
+func (r *runner) handleWCLogin(ctx context.Context) error {
+	// At the moment, the only available login option for WC is client cert
+	return r.handleWCClientCert(ctx)
+}
 
-	certOperatorVersion, err := getCertOperatorVersion(ctx, releaseService, releaseVersion)
+func (r *runner) handleWCClientCert(ctx context.Context) error {
+	var err error
+
```
tfussell, line 129:
```suggestion
```

cmd/login/wc.go
```diff
@@ -90,52 +61,138 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
...
+	config := commonconfig.New(r.flag.config)
```
tfussell, line 130:
```suggestion
	config := commonconfig.New(r.flag.config)
```

cmd/login/wc.go
```diff
@@ -90,52 +61,138 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
...
+		fmt.Fprintf(r.stdout, "  kubectl config use-context %s\n", contextName)
+	}
+
+	return nil
+}
+
+func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.Interface, provider string) (string, bool, error) {
+	var err error
+
+	err = validateProvider(provider)
```
tfussell, line 171:
```suggestion
	err := validateProvider(provider)
```
No need to declare err at the top of each function especially if you're setting it in the next lines.

cmd/login/wc.go
```diff
@@ -90,52 +61,138 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
...
+		fmt.Fprintf(r.stdout, "  kubectl config use-context %s\n", contextName)
+	} else {
+		fmt.Fprintf(r.stdout, "A new kubectl context has been created named '%s' and selected. To switch back to this context later, use this command:\n\n", contextName)
+		fmt.Fprintf(r.stdout, "  kubectl config use-context %s\n", contextName)
+	}
+
+	return nil
+}
+
+func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.Interface, provider string) (string, bool, error) {
```
tfussell, line 168: Could you name the return values or add a comment explaining what the `string` and `bool` returns indicate?

cmd/login/wc.go
tfussell: Could you add what you said above as a comment in the code?

cmd/login/wc.go
tfussell: I see what you mean. Checking for a release label and cert-operator in that release is equivalent to checking if a cluster is "pure CAPI" so I think it's fine.

cmd/login/wc.go
```diff
@@ -147,59 +204,84 @@ func (r *runner) createClusterClientCert(ctx context.Context, client k8sclient.I
...
-	} else {
-		contextName, contextExists, err = storeWCCredentials(r.k8sConfigAccess, r.fs, clientCertResource, secret, clusterBasePath, r.loginOptions)
+	}
+
+	fmt.Fprint(r.stdout, color.GreenString("\nCreated client certificate for workload cluster '%s'.\n", r.flag.WCName))
+	return contextName, contextExists, nil
+}
+
+func (r *runner) getCredentials(ctx context.Context, clientCertService clientcert.Interface, config clientCertConfig) (*clientcert.ClientCert, *v1.Secret, error) {
+	var secret *v1.Secret
```
tfussell, line 240: Can we improve the name of this variable? I initially thought it just contained the private key, but I see now that it's actually a Kubernetes secret that contains the certificate, the key, and the CA. I'd suggest `clientCertSecret` or `certSecret`.

cmd/login/clientcert.go
```diff
@@ -139,22 +157,100 @@ func fetchCredential(ctx context.Context, provider string, clientCertService cli
...
+		// Sign the certificate
+		certBytes, err := x509.CreateCertificate(rand.Reader, certificate, caCert, &certPrivKey.PublicKey, caPrivKey)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+		certPEM = getCertPEM(certBytes)
+
+	}
+	secret := &corev1.Secret{
+		Data: map[string][]byte{
```
tfussell, line 213: Should this also have `Type: "kubernetes.io/tls"` to be more standard? Not sure if Kubernetes enforces a specific structure on those kinds of secrets. See https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets

cmd/login/clientcert.go
```diff
@@ -139,22 +157,100 @@ func fetchCredential(ctx context.Context, provider string, clientCertService cli
...
+	return secret, nil
+}
+
+func getCertPEM(cert []byte) []byte {
+	return pem.EncodeToMemory(&pem.Block{
+		Type:  "CERTIFICATE",
+		Bytes: cert,
+	})
+}
+func getCert(certPEM []byte) (*x509.Certificate, error) {
```
tfussell, line 228:
```suggestion

func getCert(certPEM []byte) (*x509.Certificate, error) {
```

cmd/login/clientcert.go
```diff
@@ -139,22 +157,100 @@ func fetchCredential(ctx context.Context, provider string, clientCertService cli
...
+	block, _ := pem.Decode(certPEM)
+	return x509.ParseCertificate(block.Bytes)
+}
+func getPrivKeyPEM(key *rsa.PrivateKey) []byte {
```
tfussell, line 232:
```suggestion

func getPrivKeyPEM(key *rsa.PrivateKey) []byte {
```

cmd/login/clientcert.go
```diff
@@ -139,22 +157,100 @@ func fetchCredential(ctx context.Context, provider string, clientCertService cli
...
+	return pem.EncodeToMemory(&pem.Block{
+		Type:  "RSA PRIVATE KEY",
+		Bytes: x509.MarshalPKCS1PrivateKey(key),
+	})
+}
+func getPrivKey(keyPEM []byte) (*rsa.PrivateKey, error) {
```
tfussell, line 238:
```suggestion

func getPrivKey(keyPEM []byte) (*rsa.PrivateKey, error) {
```

internal/feature/clientcert.go
```diff
@@ -11,4 +11,7 @@ var clientCert = Feature{
 	ProviderAWS: Capability{
 		MinVersion: "16.0.1",
 	},
+	ProviderOpenStack: Capability{
+		MinVersion: "20.0.0-alpha1",
```
tfussell, line 15: We don't have any release version on OpenStack so I'm not sure if this is necessary.

cmd/login/wc_test.go
```diff
@@ -339,10 +439,18 @@ func FakeK8sClient() k8sclient.Interface {
 		if err != nil {
 			panic(err)
 		}
+		err = capz.AddToScheme(scheme)
```
tfussell, line 442: Can we just use this scheme instead of maintaining it in two places? https://github.com/giantswarm/kubectl-gs/blob/5118440802d2dfeeb947ba8d41474d2b86166216/pkg/scheme/scheme.go#L37

cmd/login/wc.go
```diff
@@ -214,5 +296,24 @@ func getClusterBasePath(k8sConfigAccess clientcmd.ConfigAccess) (string, error)
 	// Some management clusters might have 'api.g8s' as prefix (example: Viking).
 	clusterServer = strings.TrimPrefix(clusterServer, "https://api.g8s.")
 
+	// openstack clusters have an api.$INSTALLATION prefix
+	if provider == key.ProviderOpenStack {
```
tfussell, line 300: I still don't like this, but I don't think we can fix it in this PR. I have this WIP PR to improve the situation https://github.com/giantswarm/kubectl-gs/pull/713

cmd/login/wc_test.go
```diff
@@ -18,9 +24,8 @@ import (
 	"github.com/giantswarm/k8sclient/v5/pkg/k8sclient"
 	"github.com/giantswarm/k8sclient/v5/pkg/k8scrdclient"
 	"github.com/giantswarm/microerror"
-	"github.com/spf13/afero"
```
tfussell, line 27: I don't think there should be a newline here.

cmd/login/wc.go
anvddriesch: Yes, this will fail if there is no release label. We also assume that the release CR exists.
AFAIK we currently also need the label to tell if a cluster is in CAPA/CAPZ.
It worked during testing, though. Do we have MCs without releases already?

tfussell: No OpenStack MCs should have Release CRs (see https://github.com/giantswarm/openstack-app-collection/pull/39) and no WCs should have release labels (removed in cluster-openstack v0.6.0 https://github.com/giantswarm/cluster-openstack/blob/main/CHANGELOG.md#060---2022-02-28).

internal/feature/clientcert.go
tfussell: We don't have any kyverno policies on OpenStack at the moment.

cmd/login/wc.go
anvddriesch: Alright.
It's no problem to get rid of the release version concept here for openstack. I think we can also get rid of the version validation completely since all azure and aws cluster should support client cert creation at this point.
To support CAPA and CAPZ we are going to need something else however, since they are currently identified through versioning.

internal/feature/clientcert.go
anvddriesch: I got rid of the entire thing as it was not needed anymore. The only clusters in an older version we still run are 2 in `viking` and we can create clientCerts for those using the `--insecure-namespace` flag.

anvddriesch:
@tfussell since our hours barely overlap, feel free to take over this PR in my absence in case something is time sensitive.

cmd/login/clientcert.go
```diff
@@ -139,22 +156,104 @@ func fetchCredential(ctx context.Context, provider string, clientCertService cli
...
+		}
+		keyPEM = getPrivKeyPEM(certPrivKey)
+
+		// Sign the certificate
+		certBytes, err := x509.CreateCertificate(rand.Reader, certificate, caCert, &certPrivKey.PublicKey, caPrivKey)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+		certPEM = getCertPEM(certBytes)
+
```
tfussell, line 209:
```suggestion
```

tfussell (approved):
I haven't actually tested this, but the code looks good to me.

anvddriesch:
I have tested this for clusters in openstack as well as other providers to make sure nothing broke and it seems good. Let's go with this for now.

PR #735, opened 2022-04-05
Selfupdate: Synchronize changes from devctl.

Gacko:
There are changes made in devctl which could be useful for kubectl-gs.

Background: I had some trouble with generating the cache directory in opsctl and determined differences in common update routines in devctl, kubectl-gs and opsctl repository. I there would like to synchronize those changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 - Improved description of the `--control-plane-az` parameter when templating a cluster.
 
+### Fixed
+
+- Create cache directory if it does not exist, yet.
+
 ## [2.7.0] - 2022-04-01
 
 - In `kubectl gs login`, add support for workload clusters on OpenStack.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,10 +11,6 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 - Improved description of the `--control-plane-az` parameter when templating a cluster.
 
-### Fixed
-
-- Create cache directory if it does not exist, yet.
-
 ## [2.7.0] - 2022-04-01
 
 - In `kubectl gs login`, add support for workload clusters on OpenStack.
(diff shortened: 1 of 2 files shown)


pkg/selfupdate/updater.go
```diff
@@ -138,7 +140,10 @@ func (u *Updater) getLatestVersion() (semver.Version, error) {
 		u.cache.LatestVersion = latestVersion.Version.String()
 		// We ignore the error, because if saving fails, we'll just
 		// fetch the version again next time.
```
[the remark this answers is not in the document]
Gacko: Ah, yeah. Gonna remove it on the other repositories (devctl, opsctl), too.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 - Improved description of the `--control-plane-az` parameter when templating a cluster.
 
+### Fixed
+
+- Create cache directory if it does not exist, yet.
```
marians, line 16: Let's try and explain that so that every user can understand the change. I'll make an attempt and you can improve it, okay?

```suggestion
- Prevent kubectl-gs from checking the latest version on every execution, and use a cache instead.
```

marians:
I tested the latest `main` behaviour and don't really understand what this PR is trying to fix. Here (in my local test) the cache dir gets created when it doesn't exist. Did that not work for you?

CHANGELOG.md
marians: Hmm. Looks like that description doesn't really fit.

Gacko:
Sorry, this is a bit confusing. The cache directory already gets created on latest versions of kubectl-gs. So this PR is more about syncing changes from devctl. Actually the only difference was the logging part. So I guess we don't need to add a changelog entry at all. My fault, sorry!

Gacko:
The "cache directory does not get created" thing is a problem in the current opsctl version since it's even more out-of-sync with devctl than kubectl-gs.

marians:
Ok, then we can indeed get rid of the changelog entry.

marians (approved):
Approving so you are unblocked

PR #766, opened 2022-04-19
Update check: set empty API token

marians:
Towards https://github.com/giantswarm/roadmap/issues/1053

Since kubectl-gs is a public GitHub repository, there is no need to pass a token for the update check. This PR enforces an empty string instead.

An empty token also means that if the `GITHUB_TOKEN` environment variable is set, it will be used for API authentication with GitHub. This _might_ affect access to the repo's releases.

https://github.com/rhysd/go-github-selfupdate/blob/master/selfupdate/updater.go#L54-L56

### DIFF
--- a/pkg/selfupdate/updater.go
+++ b/pkg/selfupdate/updater.go
@@ -12,9 +12,6 @@ import (
 )
 
 type Config struct {
-	// GithubToken will be used when fetching versions
-	// from private GitHub repositories.
-	GithubToken string
 	// CurrentVersion is the currently installed version
 	// of the application.
 	CurrentVersion string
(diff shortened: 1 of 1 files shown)


pkg/selfupdate/updater.go
```diff
@@ -46,8 +42,7 @@ func New(c Config) (*Updater, error) {
 	var err error
 
 	u := &Updater{
-		githubToken: c.GithubToken,
-		cacheDir:    c.CacheDir,
+		cacheDir: c.CacheDir,
```
marians, line 45: Removing the `githubToken` config option, as the kubectl-gs repository is public.

anvddriesch (approved):

PR #776, opened 2022-04-25
Upgrade CAPI dependencies

anvddriesch:
Towards https://github.com/giantswarm/giantswarm/issues/21301

---


njuettner (approved):
+1 for AWS, thank you ❤️

cmd/get/clusters/printer_test.go
```diff
@@ -6,14 +6,14 @@ import (
...
+	infrastructurev1alpha3 "github.com/giantswarm/apiextensions/v6/pkg/apis/infrastructure/v1alpha3"
 	"github.com/giantswarm/k8smetadata/pkg/annotation"
 	"github.com/google/go-cmp/cmp"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
-	capzv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/api/v1alpha3"
-	capiv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
+	capz "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
+	capiv1beta1 "sigs.k8s.io/cluster-api/api/v1beta1"
```
fiunchinho, line 16: In other repos we are naming this import just `capi`, removing the version from the alias name. That way, next time we upgrade, only the import line will change and the diff is much smaller. What do you think about doing the same here?

cmd/get/clusters/printer_test.go
```diff
@@ -248,33 +248,33 @@ func Test_printOutput(t *testing.T) {
 	}
 }
 
-func newCAPIV1alpha3Cluster(id, created, release, org, description string, conditions []string) *capiv1alpha3.Cluster {
+func newcapiv1beta1Cluster(id, created, release, org, description string, conditions []string) *capiv1beta1.Cluster {
 	location, _ := time.LoadLocation("UTC")
 	parsedCreationDate, _ := time.ParseInLocation(time.RFC3339, created, location)
-	c := &capiv1alpha3.Cluster{
+	c := &capiv1beta1.Cluster{
 		TypeMeta: metav1.TypeMeta{
```
fiunchinho, line 255: is the `TypeMeta` field required? I don't think I've ever set it, and it requires being changed every time we upgrade the api version. Maybe we could remove it?

cmd/get/clusters/printer_test.go
```diff
@@ -248,33 +248,33 @@ func Test_printOutput(t *testing.T) {
 	}
 }
 
-func newCAPIV1alpha3Cluster(id, created, release, org, description string, conditions []string) *capiv1alpha3.Cluster {
+func newcapiv1beta1Cluster(id, created, release, org, description string, conditions []string) *capiv1beta1.Cluster {
```
fiunchinho, line 251: I was thinking that maybe we could rename the function name and drop the api version from its name. Something like `newCapiCluster`? WDYT?

cmd/get/nodepools/printer_test.go
```diff
@@ -7,16 +7,16 @@ import (
...
 	"github.com/giantswarm/k8smetadata/pkg/annotation"
 	"github.com/giantswarm/k8smetadata/pkg/label"
 	"github.com/google/go-cmp/cmp"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/cli-runtime/pkg/genericclioptions"
-	capzexpv1alpha3 "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1alpha3"
-	capiv1alpha3 "sigs.k8s.io/cluster-api/api/v1alpha3"
-	capiexpv1alpha3 "sigs.k8s.io/cluster-api/exp/api/v1alpha3"
+	capzexpv1beta1 "sigs.k8s.io/cluster-api-provider-azure/exp/api/v1beta1"
```
fiunchinho, line 17: same as above regarding alias name containing api version.

cmd/get/clusters/printer_test.go
anvddriesch: Sure, let's do that. :)

cmd/get/clusters/printer_test.go
anvddriesch: I'm alright with removing it but I can't tell you if it's going to break anything.

cmd/get/clusters/printer_test.go
anvddriesch: sure!

whites11 (approved):
LGTM when tested

cmd/get/nodepools/printer_test.go
```diff
@@ -354,18 +354,18 @@ func newCAPIexpv1alpha3MachinePool(name, clusterName, created, release, descript
 	parsedCreationDate, _ := time.ParseInLocation(time.RFC3339, created, location)
 	n := &capiexpv1alpha3.MachinePool{
 		TypeMeta: metav1.TypeMeta{
-			APIVersion: "exp.cluster.x-k8s.io/v1alpha3",
+			APIVersion: "exp.cluster.x-k8s.io/v1beta1",
```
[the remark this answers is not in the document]
anvddriesch: This is good to know. And it only affects machinepools?

anvddriesch:
@njuettner you'd like me to upgrade AWS as well? :slightly_smiling_face:

cmd/get/nodepools/printer_test.go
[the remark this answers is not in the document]
nprokopic: It affects CAPI MachinePools and CAPZ AzureMachinePools.

To write a bit more detailed:

MachinePools changed API group from `exp.cluster.x-k8s.io` (where it was v1alpha3) to `cluster.x-k8s.io` (where it is v1beta1 now; v1alpha4 and v1alpha3 MachinePools are also available in the new API group, but we don't and won't use these anywhere).

AzureMachinePools changed API group from `exp.infrastructure.cluster.x-k8s.io` (where it was v1alpha3) to `infrastructure.cluster.x-k8s.io` (where it is v1beta1 now; v1alpha4 and v1alpha3 MachinePools are also available in the new API group, but we don't and won't use these anywhere).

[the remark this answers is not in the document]
nprokopic: So now, for the old types (still used in all existing releases), we have old [`exp.cluster.x-k8s.io/v1alpha3` MachinePools](https://github.com/giantswarm/apiextensions/tree/v6.0.0/pkg/apis/capiexp/v1alpha3) and [`infrastructure.cluster.x-k8s.io/v1alpha3` AzureMachinePool](https://github.com/giantswarm/apiextensions/tree/v6.0.0/pkg/apis/capzexp/v1alpha3) in apiextensions, and this way we can upgrade CAPI dependencies to the latest version, and keep using old types, but just from the apiextensions now.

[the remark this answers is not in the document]
anvddriesch: So for those I'll just go with what's in apiextensions/v1alpha3 for now?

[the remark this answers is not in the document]
anvddriesch: Alright, got it!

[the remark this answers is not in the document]
nprokopic: > So for those I'll just go with what's in apiextensions/v1alpha3 for now?

Yup, you can just go with apiextensions for MachinePools for now :)

njuettner:
@anvddriesch you mean the AWS specific CRs from GiantSWarm? No, those won't be updated anymore, we keep them on `v1alpha3` 🙂.

cmd/get/clusters/printer_test.go
anvddriesch: In this test (and a couple others) we seem to need it and it breaks when removed. However, for the `template` commands it was fine to remove it and I did.
As for the `get` commands, I was under the impression that we set it explicitly to ensure the CRs will be returned in a particular API version.

fiunchinho: ok, let's keep it for now then

PR #791, opened 2022-05-12
Allow usage of any existing current context in login command

anvddriesch:
Towards https://github.com/giantswarm/roadmap/issues/1082

### DIFF
--- a/cmd/login/mc.go
+++ b/cmd/login/mc.go
@@ -13,7 +13,7 @@ import (
 )
 
 func (r *runner) handleMCLogin(ctx context.Context, installationIdentifier string) error {
-	if _, contextType := kubeconfig.IsKubeContext(installationIdentifier); contextType == kubeconfig.ContextTypeMC {
+	if _, contextType := kubeconfig.IsKubeContext(installationIdentifier); contextType == kubeconfig.ContextTypeMC || r.flag.EnforceContext {
 		return r.loginWithKubeContextName(ctx, installationIdentifier)
 
 	} else if kubeconfig.IsCodeName(installationIdentifier) {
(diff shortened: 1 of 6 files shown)


pkg/kubeconfig/auth.go
```diff
@@ -34,8 +34,11 @@ func GetAuthType(config *clientcmdapi.Config, contextName string) AuthType {
 	case authInfo.AuthProvider != nil:
 		return AuthTypeAuthProvider
 	case len(authInfo.ClientCertificate) > 0:
+		return AuthTypeClientCertificate
```
anvddriesch, line 37: Somehow the way this switch was written would return `unknowntype` immediately when a clientCertificate was present. I assume that was not the original intention so I fixed it. :see_no_evil:

cmd/login/mc.go
```diff
@@ -13,7 +13,7 @@ import (
 )
 
 func (r *runner) handleMCLogin(ctx context.Context, installationIdentifier string) error {
-	if _, contextType := kubeconfig.IsKubeContext(installationIdentifier); contextType == kubeconfig.ContextTypeMC {
+	if _, contextType := kubeconfig.IsKubeContext(installationIdentifier); contextType == kubeconfig.ContextTypeMC || r.flag.EnforceContext {
```
anvddriesch, line 16: In case the `enforce-context` flag is active we will always try to switch to a context no matter the format of the argument.

cmd/login/mc.go
```diff
@@ -83,8 +70,18 @@ func (r *runner) loginWithCodeName(ctx context.Context, codeName string) error {
 		fmt.Fprintf(r.stdout, "Switched to context '%s'.\n", contextName)
 	}
 
-	fmt.Fprint(r.stdout, color.GreenString("You are logged in to the management cluster of installation '%s'.\n", codeName))
+	return nil
+}
 
+// loginWithCodeName switches the active kubernetes context to
+// one with the name derived from the installation code name.
+func (r *runner) loginWithCodeName(ctx context.Context, codeName string) error {
```
anvddriesch, line 78: I switched the logic for `loginWithCodeName` with `loginWithKubeContextName` so that the first one calls the second and not the other way around as it used to be.

anvddriesch:
I tested this. Adding `--enforce-context` to the command will enforce switching to a context with the argument name.

kuosandys:
Tested it out, and works as expected: using the `--enforce-context` flag allows the provided context name to be treated "as is", and I can create WC client certificates with it as well 👍

cmd/login/runner.go
```diff
@@ -107,60 +108,10 @@ func (r *runner) tryToReuseExistingContext(ctx context.Context) error {
 	if err != nil {
 		return microerror.Mask(err)
 	}
-
 	currentContext := config.CurrentContext
-	kubeContextType := kubeconfig.GetKubeContextType(currentContext)
-
-	switch kubeContextType {
-	case kubeconfig.ContextTypeMC:
```
anvddriesch, line 115: I have removed the logic here and instead returned the `r.loginWithKubeContextName` function that already takes care.

kuosandys (approved):
Works well for me 🙂 Thanks for explaining the changes as well!

PR #794, opened 2022-05-17
Allow to remove cluster update schedule

njuettner:
Allow users to remove scheduled update time from cluster.
---

Issue: https://github.com/giantswarm/roadmap/issues/1074
Docs PR: https://github.com/giantswarm/docs/pull/1440

- [x] Update the public [kubectl-gs documentation](https://docs.giantswarm.io/ui-api/kubectl-gs/) to reflect the changes here.

### DIFF
--- a/cmd/update/cluster/command.go
+++ b/cmd/update/cluster/command.go
@@ -26,6 +26,7 @@ Options:
   --name <cluster-name>             	Name of the cluster to update.
   --namespace <cluster-namespace>   	Namespace of the cluster.
   --release-version <release-version>   Update the cluster to a release version. The release version must be higher than the current release version.
+  --remove-schedule						Optionally: Remove the scheduled time for the cluster.
   --scheduled-time <scheduled-time>     Optionally: Scheduled time when cluster should be updated, time format 'YYYY-MM-DD HH:MM'.
   --provider <provider> 		Name of the provider.`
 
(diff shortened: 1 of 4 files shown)


cmd/update/cluster/command.go
```diff
@@ -26,6 +26,7 @@ Options:
   --name <cluster-name>             	Name of the cluster to update.
   --namespace <cluster-namespace>   	Namespace of the cluster.
   --release-version <release-version>   Update the cluster to a release version. The release version must be higher than the current release version.
+  --remove-schedule				Optionally: Remove the scheduled time from the cluster.
```
whites11, line 29: there seems to be a spacing/alignment error before "Optionally"

whites11 (approved):

cmd/update/cluster/command.go
njuettner: thanks ❤️, fixed

whites11 (approved):

marians:
I have to say I'm not particularly fond of

- the PR and the way it does not make it easy to judge the outcome. Describing the command syntax in the PR main comment would be great.
- the fact that SIG UX hasn't been asked for review
- the short review time
- the syntax implemented:

```
kubectl gs update cluster --name abcd1 --namespace my-org --provider aws --remove-schedule
```

The semantics of the `update cluster` command so far is "make sure that a cluster gets upgraded" and I think the new function (which is about "make sure the cluster is not upgraded") deserves a separate command.

With the current implementation I would expect that using `--remove-schedule` together with other flags would at least result in an error. I don't see that happening base on the code.

njuettner:
I need to apologize here, it was not my attention not asking sig-ux. This was my mistake, I saw it as a quick win because we didn't have a way to remove the schedule for clusters without removing the annotations on Cluster CR's directly.
  

I reverted the change and I will create an issue for rainbow to have a better idea how to do it 👍🏻.

PR #812, opened 2022-06-08
Let `template cluster` add a service priority label by default

marians:
Towards https://github.com/giantswarm/giantswarm/issues/22288 https://github.com/giantswarm/giantswarm/issues/13342

This PR adds

- By default, the cluster CR output will contain the label `giantswarm.io/service-priority: highest`
- Let the user override the label value by using the `--label` or `--service-priority` flag


anvddriesch:
I added the support for `aws` vintage clusters as well since you already did `azure`. As for all the CAPI providers, there is currently no support to add labels to the cluster CR in `kgs`. (We template a configmap and while it includes labels, they will be added to all resources AFAIK).

marians:
I can't approve, as I'm the PR author, but @anvddriesch you can :)

anvddriesch (approved):

PR #813, opened 2022-06-08
Display `SERVICE PRIORITY` column in `get clusters` output

kuosandys:
Towards https://github.com/giantswarm/giantswarm/issues/22332.

This PR adds the `SERVICE PRIORITY` column to the table output of the `get clusters` command:
Screen Shot 2022-06-08 at 12 18 20

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+- Add `SERVICE PRIORITY` column to `get clusters` command table output.
+
 ### Fixed
 
 - Take `--context` flag into account when building config for `login`.
--- a/cmd/get/clusters/printer_test.go
+++ b/cmd/get/clusters/printer_test.go
@@ -15,8 +15,9 @@ import (
 	capz "sigs.k8s.io/cluster-api-provider-azure/api/v1beta1"
 	capi "sigs.k8s.io/cluster-api/api/v1beta1"
 
+	"github.com/giantswarm/k8smetadata/pkg/label"
+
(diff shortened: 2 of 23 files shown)


pipo02mix (approved):
Did not check code, but the UI seems good to me

kuosandys:
> I'm on the fence here with regard to one detail. I know we display other values like this in UPPERCASE, so it's consistent to do it for this column, too. On the other hand I hope that people understand that the actual label value has to be used in lowercase. Would it be appropriate to print that column in the original case?

That's a good point. I think printing it in the same letter casing as the expected label value would prevent adding to mental load. I'll make the changes 👍

anvddriesch:
Should we display `n/a` at all? I was personally expecting it to be `highest` when it's not set since that is the de facto priority of the cluster. Or do  we want to make clear that it has not been set?

marians:
I think it's good as it is. If we display `highest` here if the label is not set, then users will have a harder time finding clusters where the label is not used.

Giant Swarm staff should know how to interpret the label not being set.

marians (approved):

kuosandys:
> Should we display `n/a` at all? I was personally expecting it to be `highest` when it's not set since that is the de facto priority of the cluster. Or do we want to make clear that it has not been set?

I based it off of how we've currently implemented it in happa: if the service priority label is not set, we don't display the label - I guess the equivalent here is the `n/a` value. I thought it might be a bit confusing to see a value displayed that isn't set, even if `highest` service priority is assumed for handling clusters without the label, and maybe seeing the missing value will prompt users to set a service priority 🙃

anvddriesch:
Sure. It is also possible that the users don't want to treat unset clusters as if they have highest priority :smile:

anvddriesch (approved):

PR #827, opened 2022-06-09
Add service priority flag to kgs template cluster

anvddriesch:
Towards https://github.com/giantswarm/giantswarm/issues/22334

### DIFF
--- a/cmd/template/cluster/flag.go
+++ b/cmd/template/cluster/flag.go
@@ -93,6 +93,7 @@ const (
 	flagPodsCIDR                 = "pods-cidr"
 	flagRelease                  = "release"
 	flagLabel                    = "label"
+	flagServicePriority          = "service-priority"
 )
 
 type flag struct {
@@ -113,6 +114,7 @@ type flag struct {
 	BastionInstanceType      string
 	BastionReplicas          int
 	ControlPlaneInstanceType string
+	ServicePriority          string
 
 	// Provider-specific
 	AWS       provider.AWSConfig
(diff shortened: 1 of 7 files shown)


kuosandys (approved):
Tested templating clusters with the `--service-priority` flag, and it works great for both Azure and AWS 🙂 

Just checking my understanding, we don't want to allow setting this label using the `--label` flag until the Kyverno policy is in place, right?

cmd/template/cluster/flag.go
```diff
@@ -252,6 +254,7 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.PodsCIDR, flagPodsCIDR, "", "CIDR used for the pods.")
 	cmd.Flags().StringVar(&f.Release, flagRelease, "", "Workload cluster release.")
 	cmd.Flags().StringSliceVar(&f.Label, flagLabel, nil, "Workload cluster label.")
+	cmd.Flags().StringVar(&f.ServicePriority, flagServicePriority, "highest", fmt.Sprintf("Service priority of the cluster. Must be one of %v", getServicePriorities()))
```
marians, line 257: Can we use `label.ServicePriorityHighest` here instead of the string literal?

anvddriesch: Oh yeah, sure :)

anvddriesch:
> Just checking my understanding, we don't want to allow setting this label using the --label flag until the Kyverno policy is in place, right?

I think it is possible to do this, right? (Unless it is blocked by another admission controller maybe)
However, we did not want to encourage it before having the policy in place.

kuosandys:
> > Just checking my understanding, we don't want to allow setting this label using the --label flag until the Kyverno policy is in place, right?
> 
> I think it is possible to do this, right? (Unless it is blocked by another admission controller maybe) However, we did not want to encourage it before having the policy in place.

Thank you for explaining! I noticed we can't set labels with the substring `giantswarm.io` at the moment with the `--label` flag, but I wasn't sure if that was the intended effect. If I run the command with `--label="giantswarm.io/service-priority=highest` I see:
```
Error: Invalid flag: --label must contain valid label definitions (invalid label key error: "giantswarm.io/service-priority=highest": containing forbidden substring 'giantswarm.io')
```

anvddriesch:
Ah yeah I guess that is the existing validation.

PR #836, opened 2022-06-21
Adding an org

ljakimczuk:


piontec (approved):
looks good :)

kubasobon (approved):
Thanks for adding command examples

internal/gitops/structure/templates/management-cluster/template.go
```diff
@@ -0,0 +1,15 @@
+package organization
+
+import (
+	_ "embed"
+)
+
+//go:embed management-cluster.yaml.tmpl
+var managementCluster string
+
+// GetManagementClusterTemplate merges .tmpl files for organization layer.
```
uvegla, line 10:
```suggestion
// GetManagementClusterTemplate merges .tmpl files for management cluster layer.
```

internal/gitops/structure/templates/management-cluster/template.go
```diff
@@ -0,0 +1,15 @@
...
+func GetManagementClusterTemplate() []string {
```
uvegla, line 11: Just a minor thing, but I would call it `GetManagementClusterTemplates` (in plural) because it could potentially return a list of templates.

internal/gitops/structure/templates/organization/template.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	_ "embed"
+)
+
+//go:embed organization.yaml.tmpl
+var organization string
+
+// OrganizationDirectoryTemplate returns organization directory layout.
+func OrganizationDirectoryTemplate() []string {
```
uvegla, line 11: Same as for management clusters. Also maybe it is a thing in go, but I miss the verb from function names.

PR #842, opened 2022-06-23
`login`: Increase retry interval and overall timeout when creating a workload cluster client certificate

marians:
Towards https://github.com/giantswarm/giantswarm/issues/22545

Currently we wait for up to 10 seconds for a created client certificate. This PR increases this to 30. The retry interval is increased to reduce the number of requests issued in a scenario where the process is slow.


marians:
Not useful

PR #859, opened 2022-07-15
add global config flags

anvddriesch:
towards https://github.com/giantswarm/roadmap/issues/1230
This PR makes the kubectl config flags like `--context` and `--kubeconfig` global in kgs.


cmd/flag.go
```diff
@@ -10,13 +11,17 @@ const (
 )
 
 type flag struct {
+	config              genericclioptions.RESTClientGetter
```
anvddriesch, line 14: New global config flag set

cmd/get/apps/command.go
```diff
@@ -39,7 +40,7 @@ type Config struct {
 	Logger     micrologger.Logger
 	FileSystem afero.Fs
 
-	K8sConfigAccess clientcmd.ConfigAccess
+	ConfigFlags *genericclioptions.RESTClientGetter
```
anvddriesch, line 43: Instead of the old k8sconfigaccess attribute we point to the global configflags in each command config.

cmd/get/apps/command.go
```diff
@@ -65,6 +66,9 @@ func New(config Config) (*cobra.Command, error) {
 	f := &flag{}
 
 	r := &runner{
+		commonConfig: &commonconfig.CommonConfig{
```
anvddriesch, line 69: We use a `commonconfig` object in each runner since it is already used in various commands.

cmd/get/apps/flag.go
```diff
@@ -12,19 +12,16 @@ const (
 type flag struct {
 	AllNamespaces bool
 
-	config genericclioptions.RESTClientGetter
```
anvddriesch, line 15: Removing the config flags from each command since the global one is used.

cmd/login/runner_test.go
```diff
@@ -544,51 +555,6 @@ func createValidTestConfig(wcSuffix string, authProvider bool) *clientcmdapi.Con
 	return config
 }
 
-func addExtraContext(config *clientcmdapi.Config) *clientcmdapi.Config {
```
anvddriesch, line 547: Moved these to the kubeconfig test package

cmd/root.go
```diff
@@ -62,13 +56,40 @@ func New(config Config) (*cobra.Command, error) {
 
 	var err error
 
+	f := &flag{}
```
anvddriesch, line 59: We define the root command first so that we can use the pointer to the config flag in the subcommands

pkg/commonconfig/commonconfig.go
```diff
@@ -18,19 +19,23 @@ const (
 )
 
 type CommonConfig struct {
-	configFlags genericclioptions.RESTClientGetter
+	ConfigFlags *genericclioptions.RESTClientGetter
```
anvddriesch, line 22: This is changed to a pointer because we can only initialize it at runtime when the flags are parsed

kuosandys (approved):

vvondruska (approved):
Looks good to me.
Tested the flags in local environment with a few commands and both context and kubeconfig overrides worked for me

PR #864, opened 2022-08-11
capa-changes

calvix:
* remove CAPA from AWS provider - this was done based on the release version but it doesn't make sense anymore as CAPA does not have any release
* add new CAPA provider for creating capa clusters
* remove the fetching of ssh SSO public key from CP

### DIFF
--- a/.nancy-ignore
+++ b/.nancy-ignore
@@ -1,3 +1,3 @@
 # Consul - no fix yet
-CVE-2022-29153 until=2022-08-01
-CVE-2022-24687 until=2022-08-01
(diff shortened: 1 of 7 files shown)


AverageMarcus (approved):
LGTM

fiunchinho:
Do we have tests for this? I don't see changes to tests so I don't we do. We should add if we don't.

calvix:
i didn't know we have any test for kubectl-gs at all

calvix:
is that suppose to be in `runner_test.go` ?

PR #866, opened 2022-08-11
add tests for CAPA provider and fix few issues

calvix:
add test for capa providers
fix validation for capa provider
add name to the cluster configmap values

### DIFF
--- a/cmd/template/cluster/flag.go
+++ b/cmd/template/cluster/flag.go
@@ -286,6 +286,7 @@ func (f *flag) Validate() error {
 	validProviders := []string{
 		key.ProviderAWS,
 		key.ProviderAzure,
+		key.ProviderCAPA,
 		key.ProviderGCP,
 		key.ProviderOpenStack,
 		key.ProviderVSphere,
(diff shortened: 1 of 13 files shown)


cmd/template/cluster/runner_test.go
```diff
@@ -68,6 +68,37 @@ func Test_run(t *testing.T) {
 			args:               nil,
 			expectedGoldenFile: "run_template_cluster_gcp.golden",
 		},
+		{
+			name: "case 0: template cluster capa",
```
fiunchinho, line 72:
```suggestion
			name: "case 1: template cluster capa",
```

cmd/template/cluster/flag.go
```diff
@@ -438,6 +439,8 @@ func (f *flag) Validate() error {
 			// skip release validation
 		} else if f.Provider == "gcp" {
 			// skip release validation
+		} else if f.Provider == "capa" {
```
fiunchinho, line 442: [We have a function](https://github.com/giantswarm/kubectl-gs/blob/27079c8851cdb1bcc233bc50657615e2b67b38b7/internal/key/key.go#L112-L115) to detect if it's a CAPI installation. It would be better to add `capa` to that list and re-use the function.

fiunchinho (commented):
Great! Only a small suggestion.

fiunchinho (approved):

PR #876, opened 2022-08-23
Set correct gcp server address when logging in

mnitchev:
Description 
---

The GCP ControlPlaneEndpoint.Host is an IP. Since our VPN routes the domain we create for the API, this change is needed in order for us to connect the cluster

Fixes giantswarm/roadmap#1291

Checklist 
---

_I think this is N/A in this case?_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Fixed
+
+- Set domain name for the Kubernetes APIs server address when logging in to a GCP workload cluster.
+
+
 ## [2.19.2] - 2022-08-17
 
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ### Fixed
 
-- Set domain name for the Kubernetes APIs server address when logging in to a GCP workload cluster.
+- Set domain name for the Kubernetes APIs server address when logging in to CAPI provider workload clusters.
 
 
 ## [2.19.2] - 2022-08-17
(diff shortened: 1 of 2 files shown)


fiunchinho (approved):

mnitchev:
I haven't added a test here. The tests pass without me doing anything and the reason is that we currently don't really verify anything in the kubeconfig apart from the `current-context`.
The test case in [wc_test.go](https://github.com/giantswarm/kubectl-gs/blob/a757302716b43b63eb1f9679113836215855136a/cmd/login/wc_test.go#L168-L266) is very complicated (and I'm still not sure what exactly it's doing tbh) and not very easy to read or understand. I _could_ add a test to this to check the actual contents of the kubeconfig, however I fear that that  would make things even more complicated. 
I would rather we rework this test case from the ground up, but this change in particular is needed so we can connect to GCP WCs.
WDYT?

calvix:
is this only used by GCP or its for all providers?

fiunchinho:
> I haven't added a test here. The tests pass without me doing anything and the reason is that we currently don't really verify anything in the kubeconfig apart from the `current-context`. The test case in [wc_test.go](https://github.com/giantswarm/kubectl-gs/blob/a757302716b43b63eb1f9679113836215855136a/cmd/login/wc_test.go#L168-L266) is very complicated (and I'm still not sure what exactly it's doing tbh) and not very easy to read or understand. I _could_ add a test to this to check the actual contents of the kubeconfig, however I fear that that would make things even more complicated. I would rather we rework this test case from the ground up, but this change in particular is needed so we can connect to GCP WCs. WDYT?

I'm fine with reworking the test case later.

mnitchev:
>is this only used by GCP or its for all providers?
> I got the feeling Openstack/VCD might also only get Ip address and not sure how is it with the DNS

Looks like OpenStack uses IPs at least. I've asked Rocket to confirm if it's an issue for them

mnitchev:
Looks like it isn't a problem for OpenStack. I changed the comment and am merging.

PR #887, opened 2022-09-14
Remove obsolete alias in `gitops add app`

marians:
This PR removes an alias which is the same as the command name.

Before this PR, usage looks like this:

```
Aliases:
  app, app
```

This PR has no effect on the function of the command.


anvddriesch (approved):
:smile: wow, that's not even obsolete, it's just redundant.

PR #897, opened 2022-09-26
template/app: Update deprecated flag name.

Gacko:
### What does this PR do?

This PR replaces the formerly used `--cluster` flag in the help text by the new `--cluster-name` flag.

Reducing confusion, probably.

### What does it look like?

Before:
```
Usage:
  kubectl gs template app [flags]

Flags:
      [...]
      --in-cluster                      Deploy the app in the current management cluster rather than in a workload cluster. If this is set, --cluster will be ignored.
      [...]
```

After:
```
Usage:
  kubectl gs template app [flags]

Flags:
      [...]
      --in-cluster                      Deploy the app in the current management cluster rather than in a workload cluster. If this is set, --cluster-name will be ignored.
      [...]
```

None.

Probably not.

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


marians (approved):
Nice, thank you!

PR #905, opened 2022-10-04
WC name extended length

vvondruska:
### What does this PR do?

- Changes required length of resource names in validation and auto-generated name lengths of clusters, node pools and network pools to 10 characters
- Deprecates `--enable-long-names` flag from `template cluter`, `template nodepool` and `template networkpool` commands

Kgs allows users to configure longer names for resources without adding a special flag.
Auto-generated names of certain resource are longer.

### What does it look like?

Auto-generated names of Cluster, Nodepool and Neteworkpool resources are 10 characters long:
```
kind: Cluster
metadata:
  name: abc1234567
```

instead of 
```
kind: Cluster
metadata:
  name: abc12
```

Names provided manually can be shorter.

Deprecated `--enable-long-names` flag warning:

```
Flag --enable-long-names has been deprecated, Long names are supported by default, so this flag is not needed anymore and will be removed in the next major version.
```

When creating a template for (not only) a Cluster resource, Kubectl-gs fails in case a name longer than 5 characters is configured for the resource, unless a flag called `--enable-long-names` is provided.

Longer resource names are becoming more common, and since the `--enable-long-names` flag is not even visible in usage instructions, the flag is no longer needed and instead longer resource names should be supported by default.

### DIFF
--- a/cmd/template/cluster/flag.go
+++ b/cmd/template/cluster/flag.go
@@ -17,7 +17,8 @@ import (
 )
 
 const (
-	flagProvider = "provider"
+	flagEnableLongNames = "enable-long-names"
+	flagProvider        = "provider"
 
 	// AWS only.
 	flagAWSExternalSNAT       = "external-snat"
@@ -129,6 +130,7 @@ type flag struct {
 }
 
 func (f *flag) Init(cmd *cobra.Command) {
+	cmd.Flags().BoolVar(&f.EnableLongNames, flagEnableLongNames, false, "Allow long names.")
 	cmd.Flags().StringVar(&f.Provider, flagProvider, "", "Installation infrastructure provider.")
 
 	// AWS only.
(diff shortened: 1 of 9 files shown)


cmd/template/cluster/provider/templates/aws/cluster.go
```diff
@@ -54,7 +54,7 @@ func NewClusterCRs(config ClusterCRsConfig) (ClusterCRs, error) {
 	// the workload cluster name may be provided by the user.
 	{
 		if config.ClusterName == "" {
-			generatedName, err := key.GenerateName(config.EnableLongNames)
+			generatedName, err := key.GenerateName(true)
```
vvondruska, line 57: This generates a longer version of the resource name.
Can it be like this or is it better to generate 5-character names?

cmd/template/cluster/flag.go
```diff
@@ -300,7 +296,7 @@ func (f *flag) Validate() error {
 	}
 
 	if f.Name != "" {
-		valid, err := key.ValidateName(f.Name, f.EnableLongNames)
+		valid, err := key.ValidateName(f.Name, true)
```
anvddriesch, line 299: Do we ever call `ValidateName` with `false`? Otherwise we might not need the argument at all.

cmd/template/cluster/provider/templates/aws/cluster.go
```diff
@@ -63,7 +63,7 @@ func NewClusterCRs(config ClusterCRsConfig) (ClusterCRs, error) {
 		}
 
 		if config.ControlPlaneName == "" {
-			generatedName, err := key.GenerateName(config.EnableLongNames)
+			generatedName, err := key.GenerateName(true)
```
anvddriesch, line 66: Same as above

cmd/template/cluster/flag.go
vvondruska: We call it from within the GenerateName method, and pass `enableLongNames` parameter from it.

cmd/template/cluster/provider/templates/aws/cluster.go
vvondruska: We call it with `false` when generating names for catalogs. That behaviour is older than this PR, so I thought it might be safer to keep it.

anvddriesch: Alright then :) That makes sense.

anvddriesch (approved):

PR #906, opened 2022-10-05
Add option to reference 'cluster-values' configmap when creating App CRs

fiunchinho:
### What does this PR do?

We need the `default-apps-$provider` app bundle to use the `cluster-values` configmap. 

The default apps included in `default-apps-$provider` app bundle can consume values from the shared configmap.

### What does it look like?

example output

```
./kubectl-gs template cluster --context gs-gunther --provider gcp --organization giantswarm --name jos18 --description "jose test" --gcp-project giantswarm-352614 --region europe-west6 --gcp-failure-domains europe-west6-a,europe-west6-b,europe-west6-c --gcp-machine-deployment-failure-domain europe-west6-a
---
apiVersion: v1
data:
  values: |
    clusterDescription: jose test
    clusterName: jos18
    controlPlane:
      replicas: 3
      serviceAccount:
        email: default
        scopes:
        - https://www.googleapis.com/auth/compute
    gcp:
      failureDomains:
      - europe-west6-a
      - europe-west6-b
      - europe-west6-c
      project: giantswarm-352614
      region: europe-west6
    machineDeployments:
    - failureDomain: europe-west6-a
      instanceType: n1-standard-2
      name: worker0
      replicas: 3
      rootVolumeSizeGB: 100
    organization: giantswarm
kind: ConfigMap
metadata:
  creationTimestamp: null
  labels:
    giantswarm.io/cluster: jos18
  name: jos18-userconfig
  namespace: org-giantswarm
---
kind: App
metadata:
  labels:
  name: jos18
  namespace: org-giantswarm
spec:
  catalog: cluster
  config:
    configMap:
      name: ""
      namespace: ""
    secret:
      name: ""
      namespace: ""
  kubeConfig:
    context:
      name: ""
    inCluster: true
    secret:
      name: ""
      namespace: ""
  name: cluster-gcp
  namespace: org-giantswarm
  userConfig:
    configMap:
      name: jos18-userconfig
      namespace: org-giantswarm
  version: 0.28.0
---
apiVersion: v1
data:
  values: |
    clusterName: jos18
    organization: giantswarm
kind: ConfigMap
metadata:
  creationTimestamp: null
  labels:
    giantswarm.io/cluster: jos18
  name: jos18-default-apps-userconfig
  namespace: org-giantswarm
---
kind: App
metadata:
  labels:
    giantswarm.io/cluster: jos18
    giantswarm.io/managed-by: cluster
  name: jos18-default-apps
  namespace: org-giantswarm
spec:
  catalog: cluster
  config:
    configMap:
      name: jos18-cluster-values
      namespace: org-giantswarm
    secret:
      name: ""
      namespace: ""
  kubeConfig:
    context:
      name: ""
    inCluster: true
    secret:
      name: ""
      namespace: ""
  name: default-apps-gcp
  namespace: org-giantswarm
  userConfig:
    configMap:
      name: jos18-default-apps-userconfig
      namespace: org-giantswarm
  version: 0.13.2
```

The `cluster-values` configmap contains values that are related to the cluster and shared among different apps. Different apps may consume these values. We want our apps/charts to be close to upstream helm charts, so we don't want to change the value names on all apps so that they match the values inside the `cluster-values` configmap. Instead the app bundle can act as a translation layer between our `cluster-values` configmap and whatever value is required by the different apps.
For example, the `baseDomain` value that we have in the `cluster-values` configmap could be passed to the cilium app which expects the value to be called `k8sServiceHost` instead of `baseDomain`.

Do you foresee any unwanted side effects because of this change?

- [X] CHANGELOG.md has been updated (if it exists)

No

### DIFF
--- a/cmd/template/cluster/provider/capg.go
+++ b/cmd/template/cluster/provider/capg.go
@@ -183,6 +183,11 @@ func templateDefaultAppsGCP(ctx context.Context, k8sClient k8sclient.Interface, 
 			Namespace:               organizationNamespace(config.Organization),
 			Version:                 appVersion,
 			UserConfigConfigMapName: configMapName,
+			DefaultingEnabled:       false,
+			UseClusterValuesConfig:  true,
+			ExtraLabels: map[string]string{
+				k8smetadata.ManagedBy: "cluster",
+			},
 		})
 		if err != nil {
 			return microerror.Mask(err)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 ### Changed
 
 - Add `giantswarm.io/cluster` label to the 'default-apps' bundle so that it's deleted when a `Cluster` is deleted.
+- Add option to reference the `cluster-values` configmap in the `App` CR created for CAPI clusters.
 
 ## [2.23.2] - 2022-10-04
 
--- a/cmd/template/cluster/provider/capo.go
+++ b/cmd/template/cluster/provider/capo.go
@@ -7,6 +7,7 @@ import (
 	"text/template"
 
 	"github.com/giantswarm/k8sclient/v7/pkg/k8sclient"
+	k8smetadata "github.com/giantswarm/k8smetadata/pkg/label"
 	"github.com/giantswarm/microerror"
 	"sigs.k8s.io/yaml"
 
@@ -191,6 +192,11 @@ func templateDefaultAppsOpenstack(ctx context.Context, k8sClient k8sclient.Inter
 			Namespace:               fmt.Sprintf("org-%s", config.Organization),
 			Version:                 appVersion,
 			UserConfigConfigMapName: configMapName,
+			DefaultingEnabled:       false,
+			UseClusterValuesConfig:  true,
+			ExtraLabels: map[string]string{
+				k8smetadata.ManagedBy: "cluster",
+			},
 		})
 		if err != nil {
 			return microerror.Mask(err)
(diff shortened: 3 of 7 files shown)


cmd/template/cluster/provider/capg.go
```diff
@@ -182,6 +182,11 @@ func templateDefaultAppsGCP(ctx context.Context, k8sClient k8sclient.Interface,
 			Namespace:               organizationNamespace(config.Organization),
 			Version:                 appVersion,
 			UserConfigConfigMapName: configMapName,
+			DefaultingEnabled:       false,
+			UseClusterValuesConfig:  true,
+			ExtraLabels: map[string]string{
+				k8smetadata.ManagedBy: "cluster",
```
fiunchinho, line 188: This label is required so that the `app-admission-controller` doesn't block the `App` CR creation due to the `cluster-values` configmap not existing yet.

cmd/template/cluster/provider/capg.go
```diff
@@ -182,6 +182,11 @@ func templateDefaultAppsGCP(ctx context.Context, k8sClient k8sclient.Interface,
 			Namespace:               organizationNamespace(config.Organization),
 			Version:                 appVersion,
 			UserConfigConfigMapName: configMapName,
+			DefaultingEnabled:       false,
```
fiunchinho, line 185: I'm not sure what this implies, but without setting this to `false` the `config` part of the CR is removed.

nprokopic:
I see we need something like this, and this seems to get the job done, but since I am not very much familiar with kubectl-gs codebase I would like Team Rainbow to check (what Jose asked above in the description) if there are any unwanted side effects because of this change.

pipo02mix (approved):
LGTM though rainbow review would be great

PR #907, opened 2022-10-05
Add cluster label to default-apps bundle so that it's deleted when Cluster is deleted

fiunchinho:
### What does this PR do?

When a `Cluster` CR is deleted, [`cluster-apps-operator` then takes care of deleting all the `App` CRs labeled with that cluster name](https://github.com/giantswarm/cluster-apps-operator/blob/64e85e1901f8352efa25cfb31f58714ee2a1d72d/service/controller/resource/app/delete.go#L149). We want it to also delete the `default-apps-$provider` bundle.

Users no longer would have to manually delete the `default-apps-$provider` bundle when deleting a `Cluster` CR.

### What does it look like?

n/a

Up until now when we want to delete a `Cluster` we manually delete the `App` representing the cluster **and the default apps bundle**. It's not only cumbersome but users may forget.

Can you think of any downside of labeling the default apps bundle with the cluster id? any side effects?

- [X] CHANGELOG.md has been updated (if it exists)

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
+- Add `giantswarm.io/cluster` label to the 'default-apps' bundle so that it's deleted when a `Cluster` is deleted.
+
 ## [2.23.2] - 2022-10-04
 
 ### Added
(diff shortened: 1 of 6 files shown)


erkanerol:
Wow. Didn't know this functionality. I am deleting default-apps manually.

erkanerol (approved):

ljakimczuk (approved):

PR #911, opened 2022-10-11
Fix login failing to access callback server

mogottsch:
### What does this PR do?
On some OS `open.Run` blocks until the opened application (e.g. chrome) is closed again.
This leads to a situation where the callback server is not started yet and the login fails.
Replacing `Run` with `Start` leads to a non-blocking behavior. Therefore, this problem does not occur.

Fixes https://github.com/giantswarm/roadmap/issues/1500

`kubectl gs login` now works on operating systems where `open.Run` blocks.

No.

No (?)

No.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


vvondruska (approved):

marians:
@mogottsch Which OS did this happen on?

mogottsch:
> @mogottsch Which OS did this happen on?

Manjaro/Linux

PR #916, opened 2022-10-12
Add: flag for machine deployments service account email

zewolfe:
### What does this PR do?
This change adds two flags to specify the Google Cloud Provider service account used for a cluster's machine deployment and the api scopes that SA can access.

Users will be able to add the Google Cloud Provider service account and api scopes for their machine deployments using the flags `gcp-machine-deployment-sa-email` and `gcp-machine-deployment-sa-scopes` respectively.

### What does it look like?
N/A

Issue - https://github.com/giantswarm/roadmap/issues/1497

N/A

PR - https://github.com/giantswarm/docs/pull/1589/files

No


cmd/template/cluster/flag.go
```diff
@@ -162,6 +164,8 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().IntVar(&f.GCP.MachineDeployment.Replicas, flagGCPMachineDeploymentReplicas, 3, "Google Cloud Platform worker replicas")
 	cmd.Flags().StringVar(&f.GCP.MachineDeployment.FailureDomain, flagGCPMachineDeploymentFailureDomain, "europe-west6-a", "Google Cloud Platform worker failure domain")
 	cmd.Flags().IntVar(&f.GCP.MachineDeployment.RootVolumeSizeGB, flagGCPMachineDeploymentRootDiskSize, 100, "Google Cloud Platform worker root disk size")
+	cmd.Flags().StringVar(&f.GCP.MachineDeployment.ServiceAccount.Email, flagGCPMachineDeploymentServiceAccountEmail, "default", "Google Cloud Platform Service Account used by the control plane")
+	cmd.Flags().StringSliceVar(&f.GCP.MachineDeployment.ServiceAccount.Scopes, flagGCPMachineDeploymentServiceAccountScopes, []string{"https://www.googleapis.com/auth/compute"}, "Scope of the control plane's Google Cloud Platform Service Account")
```
fiunchinho, line 168: It's the Service Account used by the worker nodes on the machine deployment, not the control plane, right?

fiunchinho (approved):

vvondruska (approved):

marians:
Thanks for providing helpful PR information AND a docs PR! ❤️ 

To me the default behaviour (if the flags are not set) is not clear, and I reckon this could mean that users will wonder about the same.

If I already have a cluster, my expectation would be that the node pool would use the same SA as is used for the cluster (or rather the control plane?). Is that the case?

What exactly does `default` (the default value) mean here?

zewolfe:
Hey Marian, thanks for pointing that out :).

As it stands that is not the case. The control planes SA would require more permissions because of DNS and interactions with load balancers. There are plans to have this not be the case.

`Default` is a placeholder and indicator for where they'd need to add their SA

PR #920, opened 2022-10-17
Disable template nodepool command for capi

mogottsch:
closes https://github.com/giantswarm/roadmap/issues/1188
### What does this PR do?

This PR disables the `kubectl gs template nodepool` command when the release version specified by the `--release` flag is >= `20.0.0-alpha1`. These versions use CAPI and with CAPI node pools are configured via cluster app bundles and not custom resources. Therefore, the command does not make sense anymore.

Users will be prevented from using the `kubectl gs template nodepool` command when the associated installation uses CAPI.

### What does it look like?
```
Error: Invalid flag: --release must be < 20.0.0-alpha1, because Cluster API (CAPI) based workload clusters are not supported
```

https://github.com/giantswarm/roadmap/issues/1188

- I added tests for the cases where kubectl gs prevents the user from using the command. However, I was unable to create tests for the opposite case (where kubectl gs allows to use the command), because I do not know how to mock the `k8sclient`. If anyone knows how to do this, I would add further tests.

Already done by marian.

No, because only the command only worked for CAPZ and no customer is using CAPZ.


anvddriesch (approved):
LGTM but requested review from hydra (cloud capi team) as well :)

AverageMarcus (approved):

mogottsch:
The current implementation is incorrect. 

The command should always work when provided with the --provider aws/azure , disregarding the release version (specified by the --release flag).
It should not work with any other provider.
 
I will make these changes as soon as I have confirmation.

Issue #901, opened 2022-09-29
Kubectl-gs does not update WC client certificates in existing self.contained files


vvondruska:
When updating a client certificate in an existing self-contained file, the new certificate is not stored.

Creating a new self-contained file with the same data works fine:

Steps to reproduce

1. Generate an invalid certificate to access a workload cluster and store it in a self-contained file
```
kubectl gs login INSTALLATION --workload-cluster WC_NAME --certificate-group INVALID_GROUP --certificate-ttl 1h --self-contained /path/to/self-contained
```

2. Generate a valid certificate to access the same workload cluster and store it in the same self-contained file
```
kubectl gs login INSTALLATION --workload-cluster WC_NAME --certificate-group system:masters --certificate-ttl 1h --self-contained /path/to/self-contained
```

3. Generate the same valid certificate to access the same workload cluster and store it in a new self-contained file
```
kubectl gs login INSTALLATION --workload-cluster WC_NAME --certificate-group system:masters --certificate-ttl 1h --self-contained /path/to/new-self-contained
```

Expected behaviour:
- `kubectl --kubeconfig /path/to/self-contained get nodes` executed after creation of the first self-contained file should fail
- `kubectl --kubeconfig /path/to/self-contained get nodes` executed after the first self-contained file is updated with correct credentials should succeed and print nodes
- `kubectl --kubeconfig /path/to/new-self-contained get nodes` executed after creation of the new self-contained file should succeed and print nodes

Observed behaviour:
- `kubectl --kubeconfig /path/to/self-contained get nodes` executed after creation of the first self-contained file fails
- `kubectl --kubeconfig /path/to/self-contained get nodes` executed after the first self-contained file is updated with correct credentials fails with the same error as before the update
- `kubectl --kubeconfig /path/to/new-self-contained get nodes` executed after creation of the new self-contained file succeeds and prints nodes

PR #921, opened 2022-10-18
Fixed merging changes to existing self-contained files

vvondruska:
### What does this PR do?

It fixes a bug in storing self-contained config files for workload clusters. In case an existing entry in an existing self-contained file needs to be modified, new entry data is not stored.

This PR fixes it by removing the modified entries from the existing config file before merging the new changes. It also adds a test to ensure that modifying the self-contained config files works correctly.

No visible changes to the users, it just makes the self-contained config files more usable.

https://github.com/giantswarm/kubectl-gs/issues/901

N/A

No need

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


mogottsch (approved):
LGTM,
and could reproduce the bug from the issue and verify that this PR fixes the bug

PR #929, opened 2022-10-20
Fix for flaky tests checking resource age

vvondruska:
### What does this PR do?

Fixes flaky tests, which check for resource age in the command output.
Expected age of resources is usually 0s, but the actual age is sometimes 1s.

No impact on the users

### What does it look like?

No visible changes

**Possible cause of the flakiness:**

Output of some kubectl-gs command contains age of resources included in the output.
Therefore the tests need to check for resource age to verify that the output is correct.
The tests always work with a pre-defined set of input data - usually one or more resources of a specific kind. 
Each resource in the input data contains creation date, which is used to calculate the age of the resource.
Some resources have a specific date and time of creation, others use the current timestamp.
Creation date in the input resources is configured as string in a specific format (`2006-01-02T15:04:05Z`). 
The date string is parsed to Time object when the input resources are created from the configuration.
In case the current timestamp is used as creation date, it is formatted to string first and parsed back to Time object when used in resource creation.
This reduces accuracy of the age calculation, because dates parsed from the above format are only accurate to the nearest second.
Therefore in some cases the age may be incorrectly calculated as 1s instead of 0s, even though the input resources are returned in an output almost immediately after they are created.

The following sequence can be considered as an example:
```
2022-10-20T16:45:27.998 --> Start the test, create a resource with creation time set to the current timestamp
2022-10-20T16:45:27.999 --> Format the current timestamp to string (2022-10-20T16:45:27.999 --> "2022-10-20T16:45:27Z") and use it to create an input resource
2022-10-20T16:45:27.999 --> Input resource creation: parse the string creation time to Time object ("2022-10-20T16:45:27Z" --> 2022-10-20T16:45:27.000) and create the resource
2022-10-20T16:45:28.001 --> Run kubectl-gs command and return output with the resource
2022-10-20T16:45:28.001 --> Calculate the resource age: 2022-10-20T16:45:28.001 - 2022-10-20T16:45:27.000 = 1.001s
```


anvddriesch (approved):

marians:
Neat!

PR #936, opened 2022-10-24
Add option to reference 'cluster-values' configmap when creating CAPA App CRs

AverageMarcus:
### What does this PR do?

Implements same logic introduced in https://github.com/giantswarm/kubectl-gs/pull/906 but for CAPA clusters also.

### What does it look like?

No


AverageMarcus:
I was mistaken. This is already in place on default-apps for CAPA.

PR #941, opened 2022-11-02
Set default value of enable-long-names to true

ubergesundheit:
This PR changes the default value of the `--enable-long-names` flag to `true`. Without this, the error message for the `--name` flag validation contains the old valid length (5 chars) instead of the new 10 character limit.

image

Found this together with @Gacko 

no

no

no

Closes https://github.com/giantswarm/roadmap/issues/1590.

### DIFF
--- a/cmd/template/cluster/flag.go
+++ b/cmd/template/cluster/flag.go
@@ -130,7 +130,7 @@ type flag struct {
 }
 
 func (f *flag) Init(cmd *cobra.Command) {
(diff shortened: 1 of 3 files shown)


marians:
This fixes https://github.com/giantswarm/roadmap/issues/1590

It would be great to also deprecate the flag

Gacko:
I think the flag is already deprecated and no longer used in the code. At least we also got a deprecation warning and saw a hard-coded `true` in the `ValidateName` call instead of the value of the flag.

marians:
Oh great, then please ignore me :)

marians (approved):

PR #946, opened 2022-11-09
Change default GCP machine instance type

mnitchev:
### What does this PR do?

Use n1-standard-4 - 4cpus 15GB (see https://cloud.google.com/compute/docs/general-purpose-machines#n1_machines)

### What does it look like?

This just changes the default that ends up in the `<cluster-name>-uservalues` ConfigMap

This is needed because the `n1-standard-2` has only 2 cpus.
The nginx-ingress-controller (which is a commonly used app) has a cpu request of 2, which makes it impossible to schedule.

N/A

N/A

Yes

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ### Changed
 
-- Use `n2-standard-4` as default instance types for GAPG worker nodes.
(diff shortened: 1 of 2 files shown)


kuosandys (approved):

fiunchinho (approved):

mnitchev:
Link to docs change https://github.com/giantswarm/docs/pull/1617

PR #959, opened 2022-11-23
When creating `CertConfig` during `login` use special version `0.0.0` to make dedicated `cert-operator` instance reconcile them.

whites11:
### What does this PR do?

`cert-operator` is now running as a unique app as well as a release app.
The unique instance reconciles only `certconfig` CRs that have `cert-operator.giantswarm.io/version=0.0.0`.

This PR makes `kubectl-gs login` use `cert-operator.giantswarm.io/version=0.0.0` for legacy cert configs.
This way we have a dedicated cert-operator instance just for kubeconfigs and we can avoid timeouts.

### What does it look like?

no customer-facing change

https://github.com/giantswarm/giantswarm/issues/24722#issuecomment-1325153824
https://github.com/giantswarm/giantswarm/issues/24669

nothing special

nope

No customer facing change AFAIK.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
+- When creating `CertConfig` during `login` use special version `0.0.0` to make dedicated `cert-operator` instance reconcile them.
+
 ## [2.28.2] - 2022-11-16
 
 ## [2.28.1] - 2022-11-09
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ### Changed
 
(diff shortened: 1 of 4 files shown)


anvddriesch (approved):
Having a single cert-operator for this case sounds like the way to go.
I am not sure whether we still have clusters in `default` namespace, but in case we do, please make sure this still works.
I remember some versions of cert-operator causing trouble due to namespace.

whites11:
Thanks for the reminder about default namespace!
I checked, and since release 16.0.1 we can't run on default NS any more. Oldest cluster is on 16.1.1 now :+1:

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
+- When creating `CertConfig` during `login` use special version `0.0.0` to make dedicated `cert-operator` instance reconcile them.
```
anvddriesch, line 12:
```suggestion
- Ensure dedicated `cert-operator` version `0.0.0` is used for client certificate creation in `login` command to avoid timeouts.
```

PR #964, opened 2022-11-30
Fix `update cluster` command when scheduling an upgrade for a Cluster when the Cluster CR had no previous annotations.

whites11:
### What does this PR do?

Fixes a bug that prevented to schedule a cluster upgrade using `update cluster` when the Cluster CR had no `annotations` set.

Makes it possible to schedule a cluster upgrade using `update cluster` when the Cluster CR has no `annotations` set.

### What does it look like?

No UX changes

https://github.com/giantswarm/vodafone/issues/590

N/A

N/A

Nope

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


anvddriesch (approved):
Didn't test it, but lgtm :)

vvondruska (approved):

PR #974, opened 2022-12-08
Respect `--control-plane-instance-type` for AWS cluster templating instead of using default

AndiDog:
### What does this PR do?

Fix passing on this flag so it ends up in the manifest. Previously, the flag was ignored in favor of the default `m5.xlarge`.

The flag `--control-plane-instance-type` becomes effective for AWS cluster templating.

I'll let you decide to add the label or not. I believe it can be surprising if someone used the flag and thought it was effective.

### DIFF
--- a/cmd/template/cluster/provider/templates/aws/cluster.go
+++ b/cmd/template/cluster/provider/templates/aws/cluster.go
@@ -13,31 +13,32 @@ import (
 )
 
 const (
-	defaultMasterInstanceType = "m5.xlarge"
-	kindAWSCluster            = "AWSCluster"
-	kindAWSControlPlane       = "AWSControlPlane"
-	kindG8sControlPlane       = "G8sControlPlane"
+	defaultControlPlaneInstanceType = "m5.xlarge"
+	kindAWSCluster                  = "AWSCluster"
+	kindAWSControlPlane             = "AWSControlPlane"
+	kindG8sControlPlane             = "G8sControlPlane"
 )
 
 // +k8s:deepcopy-gen=false
 
 type ClusterCRsConfig struct {
-	ClusterName       string
-	ControlPlaneName  string
-	Credential        string
-	Domain            string
-	EnableLongNames   bool
-	ExternalSNAT      bool
-	ControlPlaneAZ    []string
-	Description       string
-	PodsCIDR          string
-	Owner             string
-	Region            string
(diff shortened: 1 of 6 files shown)


cmd/template/cluster/provider/templates/aws/cluster.go
```diff
@@ -13,31 +13,32 @@ import (
 )
 
 const (
-	defaultMasterInstanceType = "m5.xlarge"
-	kindAWSCluster            = "AWSCluster"
-	kindAWSControlPlane       = "AWSControlPlane"
-	kindG8sControlPlane       = "G8sControlPlane"
+	defaultControlPlaneInstanceType = "m5.xlarge"
```
AverageMarcus, line 16: Nice! 👍

AverageMarcus (approved):
LGTM 😄

AverageMarcus:
@AndiDog before merging do you mind double checking that the flag value is also being used correctly for CAPG/GCP clusters?

AndiDog:
Working for GCP cluster templating.

For CAPG, do I need some experimental branch for testing? →

```
Error: Invalid flag: --provider must be one of: aws, azure, capa, gcp, openstack, vsphere
```

AverageMarcus:
CAPG is GCP. I just wasn't sure which name we used here.

vvondruska (approved):

AndiDog:
> CAPG is GCP. I just wasn't sure which name we used here.

@AverageMarcus That's a little confusing, but you are right that `--provider gcp` means CAPG:

```
// PureCAPIProviders is the list of all providers which are purely based on or fully migrated to CAPI
func PureCAPIProviders() []string {
	return []string{
		ProviderCAPA,
		ProviderGCP,
		ProviderVSphere,
		ProviderOpenStack,
	}
}
```

Is that historical? For example, we never had a vintage Google integration, but only CAPG?

Anyway, what you asked is working (= CAPG), so no more obvious CLI flag parsing issues from my point of view.

AverageMarcus:
> Is that historical? For example, we never had a vintage Google integration, but only CAPG?

Correct. Not ideal but we needed a way to distinguish new vs old AWS clusters.

AndiDog:
Rebase to solve conflict. Let me know if you rather use merges in such case to ease review 😉

AverageMarcus (approved):

PR #980, opened 2022-12-09
Multiple Dex connectors in login command

vvondruska:
### What does this PR do?

Adds support to select a specific Dex from a list when running `kubectl-gs login` command
Introduces a new `--connector-id` flag, which allows the user to specify the preferred connector. When provided, the user is not prompted to select a connector from a list.

Selection of the Dex connector is an extra step in the login process. It can be overridden by specifying the new `--connector-id` flag.

The extra step and/or the flag are only needed during the first login. Maintenance of existing logins (i.e. rotation of ID and Refresh tokens) remains unchanged.

### What does it look like?

Unless the `login` command is executed with the `--connector-id` flag, Dex opens a page, which prompts the user to select a preferred connector. After that the login process continues without any additional changes.

`kubectl gs login installation-name --connector-id dex-connector-identifier`

Works towards [#1592](https://github.com/giantswarm/roadmap/issues/1592)

The changes can be tested by executing the `login` command with different inputs (new login, connector id flag, existing login) and verifying that the login process runs as expected and succeeds.

The new flag definitely needs to be documented.
And maybe the extra connector selection step can be documented somewhere as well.

Yes

Debatable

### DIFF
--- a/cmd/login/flag.go
+++ b/cmd/login/flag.go
@@ -21,6 +21,8 @@ const (
 	flagWCCertTTL           = "certificate-ttl"
 	flagSelfContained       = "self-contained"
 	flagWCInsecureNamespace = "insecure-namespace"
+
+	flagConnectorID = "connector-id"
 )
 
 type flag struct {
@@ -36,6 +38,8 @@ type flag struct {
 	WCCertTTL           string
 	SelfContained       string
 	WCInsecureNamespace bool
+
+	ConnectorID string
 }
 
 func (f *flag) Init(cmd *cobra.Command) {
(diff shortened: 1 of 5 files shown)


cmd/login/flag.go
```diff
@@ -52,6 +56,8 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.WCCertCNPrefix, flagWCCertCNPrefix, "", fmt.Sprintf(`For client certificate creation. Prefix for the name encoded in the X.509 field "CN". Requires --%s.`, flagWCName))
 	cmd.Flags().BoolVar(&f.WCInsecureNamespace, flagWCInsecureNamespace, false, fmt.Sprintf(`For client certificate creation. Allow using an insecure namespace for creating the client certificate. Requires --%s.`, flagWCName))
 
+	cmd.Flags().StringVar(&f.ConnectorID, flagConnectorID, "", "Identifier of a specific Dex connector to be used for authentication. The connector is selected during the oids authentication process by default. If this flag is specified, the selection step is skipped.")
```
kuosandys, line 59:
```suggestion
	cmd.Flags().StringVar(&f.ConnectorID, flagConnectorID, "", "Identifier of a specific Dex connector to be used for authentication. The connector is selected during the oidc authentication process by default. If this flag is specified, the selection step is skipped.")
```
small typo?

kuosandys (approved):
Tested and verified that the expected query parameters are present in the auth URL 🎉

cmd/login/flag.go
vvondruska: Fixed. Thanks for catching it.

anvddriesch (approved):
Perfect!
I deployed sandys custom dex-app to a test installation and tried to login. The connectors were filtered accordingly with and without `--cluster-admin` flag. `--connector-id` flag also worked as expected.

cmd/login/flag.go
```diff
@@ -52,6 +56,8 @@ func (f *flag) Init(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&f.WCCertCNPrefix, flagWCCertCNPrefix, "", fmt.Sprintf(`For client certificate creation. Prefix for the name encoded in the X.509 field "CN". Requires --%s.`, flagWCName))
 	cmd.Flags().BoolVar(&f.WCInsecureNamespace, flagWCInsecureNamespace, false, fmt.Sprintf(`For client certificate creation. Allow using an insecure namespace for creating the client certificate. Requires --%s.`, flagWCName))
 
+	cmd.Flags().StringVar(&f.ConnectorID, flagConnectorID, "", "Identifier of a specific Dex connector to be used for authentication. The connector is selected during the oidc authentication process by default. If this flag is specified, the selection step is skipped.")
```
marians, line 59:
```suggestion
	cmd.Flags().StringVar(&f.ConnectorID, flagConnectorID, "", "ID of the Dex connector to use for authentication. With more than one connector configured in Dex, this allows to skip the selection page.")
```

anvddriesch: It would skip either way, though, wouldn't it? Even if only one connector is there.

marians: Yep, right. Simplified it more.

marians (approved):

PR #987, opened 2022-12-16
Provider resolution improvement

vvondruska:
### What does this PR do?

It improves detection of providers in kubectl-gs login command. The providers are extracted from Kubernetes API URLs. This PR adds a new method and takes the providers from Athena responses. The original detection is kept as a fallback.

No visible impact on users

### What does it look like?

No visible changes

It is needed to handle a case when e.g. a CAPI URL would be resolved as vintage AWS because of its format.

Log in to workload clusters in installations of different providers and check that it works

No need to update docs

Debatable

No

### DIFF
--- a/pkg/data/domain/cluster/common.go
+++ b/pkg/data/domain/cluster/common.go
@@ -48,7 +48,7 @@ func (s *Service) getByName(ctx context.Context, provider, name, namespace strin
 				return nil, microerror.Mask(err)
 			}
 
-		case key.ProviderOpenStack:
+		case key.ProviderOpenStack, key.ProviderCAPA, key.ProviderKVM, key.ProviderVSphere:
 			cluster, err = s.getByNameOpenStack(ctx, name, namespace)
 			if err != nil {
 				return nil, microerror.Mask(err)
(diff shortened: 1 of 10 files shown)


calvix (approved):

pkg/data/domain/cluster/common.go
```diff
@@ -48,7 +48,7 @@ func (s *Service) getByName(ctx context.Context, provider, name, namespace strin
 				return nil, microerror.Mask(err)
 			}
 
-		case key.ProviderOpenStack:
+		case key.ProviderOpenStack, key.ProviderCAPA, key.ProviderKVM, key.ProviderVSphere:
 			cluster, err = s.getByNameOpenStack(ctx, name, namespace)
```
anvddriesch, line 52: We might want to change the name of this function to something else now

vvondruska: I agree. For nowt I wanted to implement a quick solution with a minimum impact. 
In general I think the process of provider detection and usage of data retrieved from Athena could use some refactoring, and I was thinking to rename the function then.
But it's also OK to do it now.

anvddriesch: That makes sense :) No blocker for me.

anvddriesch (approved):

PR #998, opened 2023-01-13
New provider & improved provider detection

vvondruska:
### What does this PR do?
This PR attempts to solve 2 issues:

- it adds a new provider called `cloud-director` and prevents failures while handling unknown providers in some commands (e.g. `get clusters`),

- it also introduces an improved way of detecting installation providers. In previous versions the providers were derived from the source URL supplied to kubectl-gs. In this PR the provider is retrieved primarily from Athena, and in case it fails the old method is used as a fallback.

It fixes several issues that the users faced in previous versions.

### What does it look like?

No changes in visible output.

We have a new installation, which uses a provider called `cloud-director`. Kubectl-gs was not aware of this provider and as a result some commands were failing - e.g. `login --workload-cluster` or `get clusters`. This PR adds the new provider to kubectl-gs and also improves handling unknown providers, so that the commands do not fail just because the provider is unknown (they might still fail for other reasons when dealing with new providers).

Kubectl-gs needs to detect installation provider when running specific commands (e.g. `get` or `login`). The providers used to be derived from Kubernetes API URL of the installation's management cluster (or workload cluster in some cases). The detection was very simple: if the URL has a certain structure and contains "aws" in a specific, then the provider is AWS vintage. If it contains "azure" in a specific place, then the provider is Azure Vintage. Otherwise the provider is Open Stack. Now we have more than just these 3 providers, and Kubernetes API URLs are a more heterogenous than before, so the simple detection is no longer sufficient and deriving providers from the API URLs is no longer reliable. Therefore this PR introduces a new primary way of detecting providers - they are taken from installation details pulled from Athena. In case Athena is unreachable, the old simple detection is used as a fallback.

Fixes https://github.com/giantswarm/roadmap/issues/1763

The changes can be tested by logging in to management and workload clusters in different installations and listing some resources.

There are no visible changes, so the docs do not necessarily need to be updated.

Debatable

No

### DIFF
--- a/cmd/get/clusters/printer.go
+++ b/cmd/get/clusters/printer.go
@@ -29,6 +29,8 @@ func (r *runner) printOutput(clusterResource cluster.Resource) error {
 			resource = provider.GetAzureTable(clusterResource)
 		case key.ProviderOpenStack:
 			resource = provider.GetOpenStackTable(clusterResource)
+		default:
+			resource = provider.GetCommonClusterTable(clusterResource)
 		}
 
 		printOptions := printers.PrintOptions{
--- a/cmd/login/clientcert.go
+++ b/cmd/login/clientcert.go
@@ -467,9 +467,10 @@ func getAllOrganizationNamespaces(ctx context.Context, organizationService organ
 
 func fetchCluster(ctx context.Context, clusterService cluster.Interface, provider, namespace, name string) (*cluster.Cluster, error) {
 	o := cluster.GetOptions{
-		Namespace: namespace,
-		Name:      name,
-		Provider:  provider,
+		Namespace:      namespace,
+		Name:           name,
+		Provider:       provider,
+		FallbackToCapi: true,
 	}
 
 	c, err := clusterService.Get(ctx, o)
--- a/cmd/login/login.go
+++ b/cmd/login/login.go
@@ -82,7 +82,7 @@ func (r *runner) loginWithCodeName(ctx context.Context, codeName string) error {
 // loginWithURL performs the OIDC login into an installation's
 // k8s api with a happa/k8s api URL.
 func (r *runner) loginWithURL(ctx context.Context, path string, firstLogin bool, tokenOverride string) error {
-	i, err := installation.New(ctx, path, "")
+	i, err := r.commonConfig.GetInstallation(ctx, path, "")
 	if installation.IsUnknownUrlType(err) {
 		return microerror.Maskf(unknownUrlError, "'%s' is not a valid Giant Swarm Management API URL. Please check the spelling.\nIf not sure, pass the web UI URL of the installation or the installation handle as an argument instead.", path)
 	} else if err != nil {
--- a/pkg/commonconfig/commonconfig.go
+++ b/pkg/commonconfig/commonconfig.go
@@ -21,7 +21,8 @@ const (
 )
 
 type CommonConfig struct {
-	ConfigFlags *genericclioptions.RESTClientGetter
+	ConfigFlags  *genericclioptions.RESTClientGetter
+	installation *installation.Installation
 }
 
 func New(cf genericclioptions.RESTClientGetter) *CommonConfig {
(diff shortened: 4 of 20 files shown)


cmd/get/clusters/printer.go
```diff
@@ -29,6 +29,8 @@ func (r *runner) printOutput(clusterResource cluster.Resource) error {
 			resource = provider.GetAzureTable(clusterResource)
 		case key.ProviderOpenStack:
 			resource = provider.GetOpenStackTable(clusterResource)
+		default:
+			resource = provider.GetCommonClusterTable(clusterResource)
```
vvondruska, line 33: The `get clusters` command was failing for unknown providers. Added a common cluster table rendering and applied it to all existing providers, which do not require specific handling, as well as unknown providers

cmd/login/clientcert.go
```diff
@@ -467,9 +467,10 @@ func getAllOrganizationNamespaces(ctx context.Context, organizationService organ
 
 func fetchCluster(ctx context.Context, clusterService cluster.Interface, provider, namespace, name string) (*cluster.Cluster, error) {
 	o := cluster.GetOptions{
-		Namespace: namespace,
-		Name:      name,
-		Provider:  provider,
+		Namespace:      namespace,
+		Name:           name,
+		Provider:       provider,
+		FallbackToCapi: true,
```
vvondruska, line 473: This flag indicates whether kgs should try to fetch the cluster as generic CAPI in case the cluster of the specified provider cannot be fetched. It prevents failures in commands, which depend on providers, in case the provider is unknown.

cmd/login/login.go
```diff
@@ -82,7 +82,7 @@ func (r *runner) loginWithCodeName(ctx context.Context, codeName string) error {
 // loginWithURL performs the OIDC login into an installation's
 // k8s api with a happa/k8s api URL.
 func (r *runner) loginWithURL(ctx context.Context, path string, firstLogin bool, tokenOverride string) error {
-	i, err := installation.New(ctx, path, "")
+	i, err := r.commonConfig.GetInstallation(ctx, path, "")
```
vvondruska, line 85: Instead of creating a new `Installation` instance every time it's needed, instantiate it once, store it in the common config and re-use it.

pkg/commonconfig/commonconfig.go
```diff
@@ -21,7 +21,8 @@ const (
 )
 
 type CommonConfig struct {
-	ConfigFlags *genericclioptions.RESTClientGetter
+	ConfigFlags  *genericclioptions.RESTClientGetter
+	installation *installation.Installation
```
vvondruska, line 25: Storing Installation in the common config, so that it can be loaded from Athena only once and re-used multiple times in different places of kgs.

pkg/commonconfig/commonconfig.go
```diff
@@ -36,47 +37,49 @@ func (cc *CommonConfig) GetConfigFlags() genericclioptions.RESTClientGetter {
...
+		cc.installation = i
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+	}
+	return cc.installation, nil
 }
 
-func (cc *CommonConfig) GetProviderFromInstallation(ctx context.Context, athenaUrl string, fallbackToConfig bool) (string, error) {
+func (cc *CommonConfig) GetProviderFromConfig(ctx context.Context, athenaUrl string) (string, error) {
```
vvondruska, line 58: Getting provider info from the Installation data loaded from Athena, and falling back to API URL detection in case Athena is not available.

pkg/installation/installation.go
```diff
@@ -22,10 +22,11 @@ type Installation struct {
 	Provider          string
 	Codename          string
 	CACert            string
+	SourcePath        string
```
vvondruska, line 25: Storing source path (the URL of Kubernetes API or another service provided to kgs as an argument). It is used to determine whether the Installation instance in memory can be re-used. In case an installation for a different source path is needed, it is reloaded from Athena.

pkg/installation/installation.go
```diff
@@ -22,10 +22,11 @@ type Installation struct {
...
 }
 
-func New(ctx context.Context, fromUrl string, athenaUrl string) (*Installation, error) {
-	basePath, internalApiPath, err := GetBaseAndInternalPath(fromUrl)
+func New(ctx context.Context, fromUrl, customAthenaUrl string) (*Installation, error) {
+	_, internalApiPath, athenaUrl, err := GetBaseAndInternalPath(fromUrl)
```
vvondruska, line 29: Added a possibility to provide a custom Athena URL in case it cannot be reliably derived from the Kubernetes API URL

anvddriesch (approved):

PR #999, opened 2023-01-13
ID token renewal optimization

vvondruska:
### What does this PR do?

Ensures that ID tokens needed for OIDC authentication are renewed only when they expire

Users should not be affected by this change

### What does it look like?

There are no visible changes

Kubectl-gs currently sends a request to generate a fresh ID token for OIDC every time a command is executed. Even though ID tokens eventually expire, they are meant to last longer than for the duration of a single kubectl-gs command, and therefore it is not necessary to rotate the tokens in every command execution.

Furthermore, Dex sometimes reports errors when rotating tokens saying that the renew token was claimed twice. Adjusting frequency of the token rotation may help resolve the issue or at least make it less frequent.

Fixes https://github.com/giantswarm/roadmap/issues/1732

This change can be tested by running the login command and verifying that it works when the token is fresh, as well as when it is expired.

There is no direct impact on the users, so the documentation does not need to be updated.

No

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 ### Added
 
 - Introduced a new `--connector-id` flag in the `login` command to specify a Dex connector to use and skip the selection step
+- Ensured that ID tokens needed for OIDC authentication are renewed only when they expire
 
 ## [2.30.0] - 2023-01-12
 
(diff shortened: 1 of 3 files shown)


anvddriesch (approved):
Nice :)
Might be quick to add a unit test here. These kinds of checks can break easily without noticing.

vvondruska:
You are right. Added a test.

anvddriesch (approved):
Awesome

PR #1004, opened 2023-01-19
Update CAPA subnet template types

AverageMarcus:
Blocked by https://github.com/giantswarm/cluster-aws/pull/205

### What does this PR do?

Changes the templated output of subnets for private CAPA clusters based on changes introduced in cluster-aws v0.21.0

The correct output for cluster-values for clusters going forward.

### What does it look like?

N/a

More complex config now supported in the template which caused it to change from a string array to an object array

No 

No

### DIFF
--- a/cmd/template/cluster/provider/capa.go
+++ b/cmd/template/cluster/provider/capa.go
@@ -107,9 +107,16 @@ func templateClusterAWS(ctx context.Context, k8sClient k8sclient.Interface, outp
 				return microerror.Mask(err)
 			}
 
+			flagValues.Network.Subnets = []capa.Subnet{
+				{
+					CidrBlocks: []capa.CIDRBlock{},
+				},
+			}
+
 			for i := 0; i < subnetCount; i++ {
-				flagValues.Network.Subnets = append(flagValues.Network.Subnets, capa.Subnet{
-					CidrBlock: subnets[i].CIDR().String(),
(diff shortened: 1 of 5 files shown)


cmd/template/cluster/provider/capa.go
```diff
@@ -109,7 +109,12 @@ func templateClusterAWS(ctx context.Context, k8sClient k8sclient.Interface, outp
 
 			for i := 0; i < subnetCount; i++ {
 				flagValues.Network.Subnets = append(flagValues.Network.Subnets, capa.Subnet{
-					CidrBlock: subnets[i].CIDR().String(),
+					CidrBlocks: []capa.CIDRBlock{
+						{
+							CIDR:             subnets[i].CIDR().String(),
+							AvailabilityZone: fmt.Sprintf(string('a' - 1 + (1 + i))), // generate `a`, `b`, etc. based on which index we're at
```
calvix, line 115: black magic

calvix (approved):

cmd/template/cluster/provider/capa.go
calvix: what is teh point of removing 1 and then adding it? :p

calvix: `mt.Sprintf(string('a'  + i))` is that enought?

AverageMarcus: Yeah, you're right. I copy&paste from internet before going to get lunch :p 

Just updated it now and hopefully fixed lint complaints too

fiunchinho (approved):

PR #1027, opened 2023-02-22
CAPA: Rename hidden parameter `--role` to `--aws-cluster-role-identity-name` and adapt manifest output to the new name `awsClusterRoleIdentityName`

AndiDog:
### What does this PR do?

Adapt to new Helm template key name (see https://github.com/giantswarm/cluster-aws/pull/192)

Likely none, since the flag is still hidden and the use case is mostly on GS creating management clusters for customers who want to create clusters in another AWS account (by point the `AWSClusterRoleIdentity` to another account).

### What does it look like?

```yaml
---
apiVersion: v1
data:
  values: |
    aws:
      awsClusterRoleIdentityName: default2
    # [...]
kind: ConfigMap
metadata:
  creationTimestamp: null
  labels:
    giantswarm.io/cluster: kubectlgs
  name: kubectlgs-userconfig
  namespace: org-giantswarm
```

[(Please link public issues or summarize if not public.)](https://github.com/giantswarm/cluster-aws/pull/192)

Yes, but only for a hidden flag so I didn't mark it


calvix (approved):

anvddriesch (commented):
This is a breaking change. Is kgs already used for cluster creation with CAPA? Despite being a hidden flag, I would recommend deprecating the old flag but allowing both.

fiunchinho:
We only have one customer in CAPA atm, and creating workload clusters for this customer is a very limited process, which is assisted by us (I don't think they even run `kubectl-gs`).

anvddriesch (approved):
Oh alright, makes sense then :)

PR #1044, opened 2023-03-29
`kubectl gs template cluster`: name default node pool `nodepool0` instead of `machine-pool0`

marians:
### What does this PR do?

Fixes https://github.com/giantswarm/roadmap/issues/2244

The `kubectl gs template cluster` command for `--provider capa` (Cluster API provider AWS) creates a cluster with a default node pool that is named `machine-pool0`. As of the [cluster-aws v0.24.0 schema](https://github.com/giantswarm/cluster-aws/blob/v0.24.0/helm/cluster-aws/values.schema.json#L110), this values is invalid.

Also the `--machine-pool-name` flag has the same invalid default value.

This PR changes the default value to `nodepool0`.

If not specified differently, new created clusters will have a node pool named `nodepool0`.

Potentially yes, haven't checked yet.

No, I don't consider this a breaking change.


anvddriesch (approved):
Hydra might have comments but looks good to me :)

calvix (approved):
LGTM, only small note is that I noticed we have different names for default worker pool compared to `cluster-aws` where its https://github.com/giantswarm/cluster-aws/blob/master/helm/cluster-aws/values.yaml#L51

marians:
> LGTM, only small note is that I noticed we have different names for default worker pool compared to `cluster-aws` where its https://github.com/giantswarm/cluster-aws/blob/master/helm/cluster-aws/values.yaml#L51

Right. Didn't notice this before, however I don't see a problem here.

PR #1078, opened 2023-06-02
Marian's dependency updates

marians:
This is mainly a test of nancy-fixer.

I updated a few dependencies manually to see if we can get rid of .nancy-ignore entirely.


marians:
```
# github.com/giantswarm/kubectl-gs/v2/pkg/oidc
pkg/oidc/oidc.go:65:17: literal copies lock value from *provider: github.com/coreos/go-oidc/v3/oidc.Provider contains sync.Mutex
```

PR #1038, opened 2023-03-15
initial capz template

bavarianbidi:
towards https://github.com/giantswarm/roadmap/issues/2115 and https://github.com/giantswarm/giantswarm/issues/24603

(Please set a descriptive PR title. Use this space for additional explanations.)

#### valid/working template generation

```
./kubectl-gs-linux template cluster --provider=capz --name=mario02 --cluster-config=out/capz_cluster.yaml --default-app-config=out/capz_default_app.yaml --kubeconfig=/home/marioc/glippy.kubeconfig
```

```yaml
# out/capz_cluster.yaml 
metadata:
  name: "cluster01"
  description: "cluster01 test cluster"
  organization: "multi-project"
providerSpecific:
  location: "westeurope"
  #This is the GHOST subscription
  subscriptionId: 6b1f6e4a-6d0e-4aa4-9a5a-fbaca65a23b3
  azureClusterIdentity:
    name: "cluster-identity-static-sp"
nodePools:
  - name: md00
    instanceType: Standard_D2s_v5
    replicas: 3
    rootVolumeSizeGB: 50
```

```yaml
# out/capz_default_app.yaml 
clusterName: "cluster01"
organization: "multi-project"
```

##### generated output
<details>
  <summary>generated manifest used for successful cluster creation</summary>

```yaml
---
apiVersion: v1
data:
  values: |
    metadata:
      description: mario02 cluster
      name: mario02
      organization: multi-project
    nodePools:
    - instanceType: Standard_D2s_v5
      name: md00
      replicas: 3
      rootVolumeSizeGB: 50
    providerSpecific:
      azureClusterIdentity:
        name: cluster-identity-static-sp
      location: westeurope
      subscriptionId: 6b1f6e4a-6d0e-4aa4-9a5a-fbaca65a23b3
kind: ConfigMap
metadata:
  creationTimestamp: null
  name: mario02-user-values
  namespace: org-multi-project
---
apiVersion: application.giantswarm.io/v1alpha1
kind: App
metadata:
  labels:
    app-operator.giantswarm.io/version: 0.0.0
  name: mario02
  namespace: org-multi-project
spec:
  catalog: cluster
  config:
    configMap:
      name: ""
      namespace: ""
    secret:
      name: ""
      namespace: ""
  kubeConfig:
    context:
      name: ""
    inCluster: true
    secret:
      name: ""
      namespace: ""
  name: cluster-azure
  namespace: org-multi-project
  userConfig:
    configMap:
      name: mario02-user-values
[...]

PR #1082, opened 2023-06-13
Add minimal support for templating CAPZ clusters

AndiDog:
### What does this PR do?

Towards https://github.com/giantswarm/roadmap/issues/525

None, since cluster templating for CAPZ wasn't a supported feature before

### What does it look like?

This minimal command now works:

```
kubectl gs template cluster --provider capz --organization testing --name bam --region westeurope --azure-subscription-id 00000000-0000-0000-0000-000000000000
```

Mainly to make our "Create a workload cluster" documentation consistent, since I'm currently working on that (https://github.com/giantswarm/roadmap/issues/2393)

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
+- Add minimal support for templating CAPZ clusters
+
 ### Changed
 
 - App: Rename `nginx-ingress-controller-app` to `ingress-nginx`. ([#1077](https://github.com/giantswarm/kubectl-gs/pull/1077))
@@ -35,7 +39,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ### Changed
 
-- Stop using old `v1alpha3` version when using CAPI CRDs. 
+- Stop using old `v1alpha3` version when using CAPI CRDs.
 
 ## [2.34.1] - 2023-03-30
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,9 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
-### Added
+### Breaking changes
 
-- Add minimal support for templating CAPZ clusters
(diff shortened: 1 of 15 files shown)


fiunchinho (approved):

bavarianbidi (changes requested):
Coming from https://gigantic.slack.com/archives/C02FJACQ3D4/p1678256339790879
in https://github.com/giantswarm/kubectl-gs/pull/1038 the `CAPZ` template feature got implemented in `kubectl-gs` (public doc: https://docs.giantswarm.io/getting-started/capz-cluster-creation/)

The main reason for doing the implementation that way was not having a second place where all the possible flags has to be maintained.

Maybe something which has to be discussed again?!

AndiDog:
@bavarianbidi (and ex @giantswarm/team-clippy) we discussed in KaaS Sync that anyhow a new command `kubectl gs create cluster` will totally replace the whole manually-crafted code. With that, we said that merging _this here_ for the short term is fine, since we won't work on it anymore and at least I can make the commands in our documentation the same for now. Let me know if you have concerns or a LGTM.

uvegla:
If not too much trouble, could you please hook it into `gitops add base` command: https://github.com/giantswarm/kubectl-gs/blob/be802a0f2213f69266645e69a96d02580e16a03b/cmd/gitops/add/base/runner.go#L92-L103 🙏🏻 

If the conventions were followed it should be fairly straightforward.

bavarianbidi:
> > The main reason for doing the implementation that way was not having a second place where all the possible flags has to be maintained.
> > Maybe something which has to be discussed again?!
> 
> I had the same feeling. Too many ways to template stuff, and especially all the CLI params are not maintainable. Let me bring this up in KaaS sync, maybe.

sure, i wont block this but then let's drop the `template` implementation again (introduced in https://github.com/giantswarm/kubectl-gs/pull/1038) + remove the existing public documentation https://docs.giantswarm.io/getting-started/capz-cluster-creation/ as well.

AndiDog:
Thanks for the input 👍. I've implemented the suggestions.

@uvegla `kubectl gs gitops add base --provider capz --region westeurope --azure-subscription-id BLABLA` now works.

@bavarianbidi `--cluster-config`/`--default-app-config` templating option code is removed.

bavarianbidi (approved):

PR #1102, opened 2023-08-09
login-add-eks-cluster-support

calvix:
### What does this PR do?

(Please set a descriptive PR title. Use this space for additional explanations.)

### What does it look like?

- [ ] CHANGELOG.md has been updated

### DIFF
--- a/cmd/login/aws.go
+++ b/cmd/login/aws.go
@@ -1,13 +1,17 @@
 package login
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/giantswarm/k8sclient/v7/pkg/k8sclient"
 	"github.com/giantswarm/microerror"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/client-go/tools/clientcmd"
 	clientcmdapi "k8s.io/client-go/tools/clientcmd/api"
+	eks "sigs.k8s.io/cluster-api-provider-aws/controlplane/eks/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/yaml"
 
 	"github.com/giantswarm/kubectl-gs/v2/pkg/kubeconfig"
 )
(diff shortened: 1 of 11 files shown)


cmd/login/aws.go
```diff
@@ -0,0 +1,158 @@
...
+		// Create authenticated user.
+		user, exists := config.AuthInfos[userName]
+		if !exists {
+			user = clientcmdapi.NewAuthInfo()
+		}
+
+		user.Exec = &clientcmdapi.ExecConfig{
+			APIVersion: "client.authentication.k8s.io/v1beta1",
+			Command:    "aws",
+			Args:       []string{"--region", c.region, "eks", "get-token", "--cluster-name", c.clusterName, "--output", "json"},
```
calvix, line 56: this is copied from how it looks like when you create kubeconfig via `aws eks update-kubeconfig`

cmd/login/clientcert.go
```diff
@@ -269,7 +269,7 @@ func getPrivKey(keyPEM []byte) (*rsa.PrivateKey, error) {
 }
 
 // storeWCClientCertCredentials saves the created client certificate credentials into the kubectl config.
-func storeWCClientCertCredentials(k8sConfigAccess clientcmd.ConfigAccess, fs afero.Fs, c credentialConfig, mcContextName string) (string, bool, error) {
+func storeWCClientCertCredentials(k8sConfigAccess clientcmd.ConfigAccess, c credentialConfig, mcContextName string) (string, bool, error) {
```
calvix, line 272: removing `fs afero.Fs` as it is never used

cmd/login/runner.go
```diff
@@ -29,13 +29,13 @@ type runner struct {
...
-	switchToClientCertContext bool
-	originContext             string
-	contextOverride           string
+	selfContained     bool
+	selfContainedWC   bool
+	isWC              bool
+	switchToContext   bool
+	switchToWCContext bool
+	originContext     string
+	contextOverride   string
```
calvix, line 38: the naming here was misleading so i corrected the flags to what they are really used for

calvix: `isWCClientCert` ->`isWC`
`selfContainedClientCert` -> `selfContainedWC`|
`switchToClientCertContext` - > `switchToWCContext`

marians:
Pinged bigmac as the owners of the login command

cmd/login/runner.go
anvddriesch: I think we should keep the `clientCert` within the name since there is also the case of OIDC on a workload cluster which would not require these options.

cmd/login/runner.go
```diff
@@ -100,9 +100,11 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		return microerror.Maskf(invalidConfigError, "Invalid number of arguments.")
 	}
 
-	// Clientcert creation if desired
-	if r.loginOptions.isWCClientCert {
-		return r.handleWCClientCert(ctx)
+	if r.loginOptions.isWC {
+		err := r.handleWCKubeconfig(ctx)
```
anvddriesch, line 104: Same here, the function being called is specifically for handling client certs on a workload cluster. (We are planning to move this to a different command so it'll be helpful to keep the name)

cmd/login/runner.go
calvix: But none of the options are actually tied to a clientCert vs OIDC decision-making, they are purely decision-making between WC and MC

calvix: So even with OIDC WC these are valid options: 

`selfContainedWC` self-contained kubeconfig is still a valid  option for both OIDC or clientcert 
`switchToWCContext`  switching to context or not is still valid for both OIDC or client cert


and ATM `kubectl gs` do not have any kind of implementation to support OIDC for WC login, it can only login into MC with OIDC as it needs athena on the cluster

cmd/login/runner.go
calvix: the function for handling clientcert is still there it just moved later to avoid code duplication
https://github.com/giantswarm/kubectl-gs/pull/1102/files#diff-1f160945410f9311805f9b1ff5284e4cbb59cdd8934057db5bd7b3e40049979fR199-R271

cmd/login/runner.go
anvddriesch: I get where you are coming from since it's not the default but we actually have some customers that use kgs to log into their WCs via OIDC.
The difference is that in case of WC client-cert we go the following route:
log into MC via OIDC to get resources from MC -> log into WC via client-cert
So there is an extra context switch in the middle we need to be mindful of in case of self-contained contexts and keeping the initial context selected.
Functionally nothing changes by naming the options differently, but by equating this to `WC` case we make the code even harder to read than it already is since in some cases we log into a WC and yet the option will be false.

In case of WC OIDC we do not need to log into the MC first since we do not need any resources from there and can log into the WC directly. It is kinda the same as logging into MC other than that the flags might contain both MC name + WC name for selecting this context. 

It seems that the EKS case is kind of in the middle. While we do need to log into the MC first, we are not creating a client certificate. :thinking: So I guess you are right in that the name does not apply. I can't think of a great name but, how about `secondContext`?

calvix:
Added few comments as we agreed in the call.

cmd/login/runner.go
calvix: resolved in call

PR #1124, opened 2023-10-06
Support deletion prevention for `kubectl gs template app` command

AndiDog:
### What does this PR do?

https://github.com/giantswarm/giantswarm/issues/28310

Support [deletion prevention](https://docs.giantswarm.io/advanced/deletion-prevention/) through a new flag.

This is both helpful for customers and our own use (e.g. in MC bootstrapping where we typically don't want an MC to be deletable).

### What does it look like?

```
$ kubectl-gs template app --catalog=giantswarm --cluster-name="anditest" --name=nginx-ingress-controller-app --target-namespace=kube-system --version=2.30.0  --user-configmap <(echo "controller: { service: { public: false } }" | yq -y) --prevent-deletion
---
apiVersion: v1
data:
  values: |
    controller:
      service:
        public: false
kind: ConfigMap
metadata:
  creationTimestamp: null
  labels:
    giantswarm.io/prevent-deletion: "true"
  name: nginx-ingress-controller-app-userconfig-anditest
  namespace: anditest
---
kind: App
metadata:
  labels:
    giantswarm.io/prevent-deletion: "true"
  name: nginx-ingress-controller-app
  namespace: anditest
spec:
  catalog: giantswarm
  kubeConfig:
    inCluster: false
  name: nginx-ingress-controller-app
  namespace: kube-system
  userConfig:
    configMap:
      name: nginx-ingress-controller-app-userconfig-anditest
      namespace: anditest
  version: 2.30.0
```

Yes, the article linked above.

No. The flag defaults to false.

### DIFF
--- a/cmd/login/runner_config_modify_test.go
+++ b/cmd/login/runner_config_modify_test.go
@@ -268,6 +268,7 @@ func TestKubeConfigModification(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
+		tc := tc
 		t.Run(tc.name, func(t *testing.T) {
 			configDir, err := os.MkdirTemp("", "loginTest")
 			if err != nil {
--- a/cmd/template/app/runner.go
+++ b/cmd/template/app/runner.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"text/template"
 
+	"github.com/giantswarm/k8smetadata/pkg/label"
 	"github.com/giantswarm/microerror"
 	"github.com/giantswarm/micrologger"
 	"github.com/spf13/cobra"
@@ -115,7 +116,7 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 			if userSecret.Labels == nil {
 				userSecret.Labels = map[string]string{}
 			}
-			userSecret.Labels["giantswarm.io/prevent-deletion"] = "true" //nolint:goconst
+			userSecret.Labels[label.PreventDeletion] = "true" //nolint:goconst
 		}
 
 		userConfigSecretYaml, err = yaml.Marshal(userSecret)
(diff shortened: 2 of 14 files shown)


cmd/template/app/runner.go
```diff
@@ -147,6 +162,10 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 	}
 	appConfig.NamespaceConfigLabels = namespaceLabels
 
+	if r.flag.PreventDeletion {
+		appConfig.ExtraLabels["giantswarm.io/prevent-deletion"] = "true"
```
whites11, line 166: how about extracting the label to a constant?

cmd/login/runner_config_modify_test.go
```diff
@@ -268,6 +268,7 @@ func TestKubeConfigModification(t *testing.T) {
 	}
 
 	for _, tc := range testCases {
+		tc := tc
```
fiunchinho, line 271: How come is this now necessary?

fiunchinho:
How are CAPI/CAPA/CAPZ/etc CRs protected? I was expecting we passed the value to the cluster-app so that it added the prevention label.

cmd/login/runner_config_modify_test.go
AndiDog: Lint error of implicit memory aliasing. Newer Go versions finally get rid of this by making the loop variables locally-scoped, but for now we need this. My changes didn't lead to the build errors, but I wanted to fix them.

cmd/template/app/runner.go
AndiDog: Right. Let's add it in k8smetadata: https://github.com/giantswarm/k8smetadata/pull/83

AndiDog:
> How are CAPI/CAPA/CAPZ/etc CRs protected? I was expecting we passed the value to the cluster-app so that it added the prevention label.

That needs to go into `kubectl gs template cluster`, and I think we should have that as well. Team Turtles wants to move shared stuff into a central cluster chart, so we may wait for that before going on with https://github.com/giantswarm/giantswarm/issues/28310. We can't just add the label to all resources in the `cluster-{aws,...}` chart since some need to be deletable on Helm upgrade (e.g. `name: mycluster-<hash>`).

I don't want `kubectl gs template app --prevent-deletion` to set some magic `.Values.metadata.preventDeletion=true` implicitly, but really only protect the `App`.

fiunchinho (approved):

mogottsch (approved):
Nice, thanks for making the prevent deletion mechanism more usable!

AndiDog:
`go mod tidy` was missing. Auto-merging afterwards.

PR #1125, opened 2023-10-10
Listen only on localhost for callbacks

AndiDog:
### What does this PR do?

Not listen on all addresses anymore, since we only use `http://localhost` as callback URL.

This avoids macOS firewall popups for any new kubectl-gs binary (also during `make test`). We also don't unnecessarily leave a port open anymore.

None. Cluster login still works.

### What does it look like?

_Not_ like this anymore:

Firewall popup

This annoyed me for a while 😂

No

No


ubergesundheit:
Can we make this configurable? Use case would be running kubectl-gs in a container.

AndiDog:
> Can we make this configurable? Use case would be running kubectl-gs in a container.

Makes sense. I added and tested `kubectl gs login --callback-host 0.0.0.0` (you can also use an empty value instead of `0.0.0.0`).

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Fixed
+
+- Listen only on localhost for callbacks
```
marians, line 12:
```suggestion
- `kubectl-gs login`: listen only on localhost for callbacks
```

marians (approved):

PR #1179, opened 2023-11-08
Add CAPA cluster templating parameter `--control-plane-load-balancer-ingress-allow-cidr-block` which automatically adds NAT Gateway IPs of the MC to the allowlist

AndiDog:
### What does this PR do?

Implements https://github.com/giantswarm/roadmap/issues/2883

New parameters for CAPA cluster templating

### What does it look like?

```sh
$ kubectl-gs template cluster --provider capa --name kgstest --organization giantswarm --control-plane-load-balancer-ingress-allow-cidr-block "16.0.0.0/8" --management-cluster golem | grep -A6 controlPlane:
    controlPlane:
      loadBalancerIngressAllowCidrBlocks:
      - 3.11.43.118/32
      - 35.177.163.198/32
      - 35.177.19.14/32
      - 16.0.0.0/8
    metadata:
```

I will adapt https://docs.giantswarm.io/advanced/cluster-management/private-clusters/#kubernetes-api-allowlist.

no

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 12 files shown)


whites11 (approved):
seems legit

PR #1185, opened 2023-11-09
Remove unused and outdated `IsCAPIVersion` function

AndiDog:
Dead code, no changelog needed

### DIFF
--- a/internal/key/key.go
+++ b/internal/key/key.go
@@ -95,21 +95,6 @@ func GenerateAssetName(values ...string) string {
 	return strings.Join(values, "-")
 }
 
-// IsCAPIVersion returns whether a given GS Release Version uses the CAPI projects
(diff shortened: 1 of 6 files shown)


internal/key/key.go
```diff
@@ -95,21 +95,6 @@ func GenerateAssetName(values ...string) string {
 	return strings.Join(values, "-")
 }
 
-// IsCAPIVersion returns whether a given GS Release Version uses the CAPI projects
-func IsCAPIVersion(version string) (bool, error) {
-	capiVersion, err := semver.New(FirstCAPIRelease)
```
calvix, line 100: should we also remove `FirstCAPIRelease` const?

AndiDog: done

fiunchinho (approved):

calvix (approved):

PR #1210, opened 2023-11-28
default generated name to 10 chars

ssyno:
### What does this PR do?

Changes the amount of characters used on randomly generated cluster names to 10.

### DIFF
--- a/internal/key/key.go
+++ b/internal/key/key.go
@@ -27,7 +27,8 @@ const (
 	// (does not contain 1 and l, to avoid confusion)
 	NameChars = "023456789abcdefghijkmnopqrstuvwxyz"
 	// NameLengthLong represents the number of characters used to create a resource name when --enable-long-names feature flag is used.
-	NameLengthLong = 20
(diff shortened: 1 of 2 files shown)


internal/key/key.go
```diff
@@ -27,7 +27,8 @@ const (
 	// (does not contain 1 and l, to avoid confusion)
 	NameChars = "023456789abcdefghijkmnopqrstuvwxyz"
 	// NameLengthLong represents the number of characters used to create a resource name when --enable-long-names feature flag is used.
```
gawertm, line 29: i think the --enable-long-names flag is deprecated anyways. maybe we can remove references to it? and also call the var then `NameLengthMax` ? or are there too many references to it?

ssyno: the --enable-long-names is deprecated and it is enabled by default. So the existing name lengths are all defaulting to NameLengthMax

anvddriesch (approved):
please add a changelog entry

PR #1293, opened 2024-03-26
Fix authentication failure

vvondruska:
### What does this PR do?

Fix authentication failure in case the browser sends multiple requests to the callback server during the `login` command execution

The user experiencing problems with the `login` command resulting in the "SSO authentication failed" page being displayed in the browser will not experience the problem anymore

### What does it look like?

No visible changes apart

Some users experienced problems when logging in to clusters using `kubectl gs login` standard flow. The command did not finish successfully. Instead the users saw a mysterious "SSO authentication failed" page loaded in their browser.

The problem was observed in specific environments - either certain Linux distributions or certain DEV/Canary versions of the browser (e.g. MS Edge Canary).

The root cause is that kubectl-gs expects the browser to make one GET request to the local callback server started during the login command, but browsers in some cases/environments send multiple requests (e.g. an OPTIONS request in addition to the expected GET request), and the callback server was not built to handle the additional requests.

The best way to test the changes us to pull the latest changes from the PR, build kubectl-gs locally, use it to run the `login` command and check whether it works.

A Dev browser like MS Edge Canary can be used to test it in conditions, where it was failing.

No need to update the docs.

No

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 ### Changed
 
 - Errors during update checks no longer interrupt the command execution.
+- Fix authentication failure in case the browser sends multiple requests to the callback server during the `login` command execution
 
 ## [2.52.1] - 2024-02-01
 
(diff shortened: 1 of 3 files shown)


anvddriesch (approved):
lgtm

PR #1332, opened 2024-05-28
Allow app update to version from any catalog

TheoBrigitte:
### What does this PR do?

This PR add the capability to update an app to any version from any catalog.

This is achieved by matching the app name and version to a catalog entry from any catalog. This means an app deployed with version 1.0.0 from catalog A can be updated to version 2.0.0 from catalog B.

This PR also adds a `--suspend` flag to handle Flux app reconciliation. This flag is used to suspend the reconciliation of the app when the update is applied. This is useful when the app managed by Flux.

Users can now update an app to any version from any catalog.
It also allows GiantSwarm engineers to deploy app test version onto management clusters.

### What does it look like?

Example:
```
$ kubectl gs update app --name prometheus-meta-operator --namespace giantswarm --version 4.75.1-eb3c621412224384484b9d2f6f9cb952040987a5 --suspend
App 'prometheus-meta-operator' updated to version '4.75.1-eb3c621412224384484b9d2f6f9cb952040987a5'

$ kubectl --namespace giantswarm get app prometheus-meta-operator -ojson | jq -r '{annotations:.metadata.annotations,catalog:.spec.catalog,version:.spec.version}'
{
  "annotations": {
    "kustomize.toolkit.fluxcd.io/reconcile": "disabled"
  },
  "catalog": "control-plane-test-catalog",
  "version": "4.75.1-eb3c621412224384484b9d2f6f9cb952040987a5"
}
```

Might want to refresh this piece of documentation: https://docs.giantswarm.io/vintage/use-the-api/kubectl-gs/update-app/

Interface is not broken (only adding 1 argument), but default behavior changes.


marians:
The presence of the `--suspend` flag IMO suggests that the command can now be used simply to turn Flux suspension on or off. And that would be great.

```
kubectl-gs update app --namespace ns --name foobar --suspend=true
kubectl-gs update app --namespace ns --name foobar --suspend=false
```

But the implementation doesn't allow this. The `--version` flag is required, which doesn't make sense with the use case. Also, if I understand the code correctly, the execution of the patch function relies on the condition `len(options.Version) > 0`. I think this should be changed to leave the version untouched.

However, if this use case should not be supported, it should be made clear in the flag description that `--suspend` must be used with a `--version` specified.

Regarding the naming `--suspend`, I think this is a bit too general. I know, someone would love to type fewer letters, but that's what aliases and shells scripts are good for. How about `--suspend-reconciliation`.

What happens if the App CR is not reconciled by Flux (not a GitOps-managed resource) and the user uses the command with a version and `--suspend=true`? I think ideally the user should be at least informed in that case.

pkg/data/domain/app/service.go
```diff
@@ -118,21 +119,36 @@ func (s *Service) patchVersion(ctx context.Context, namespace string, name strin
...
+	}
+	annotations := accessor.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string)
+	}
+	if suspend {
+		annotations["kustomize.toolkit.fluxcd.io/reconcile"] = "disabled"
+	} else {
+		delete(annotations, "kustomize.toolkit.fluxcd.io/reconcile")
+	}
```
marians, line 150: We should have the strings `"kustomize.toolkit.fluxcd.io/reconcile"` and `"disabled"` in https://github.com/giantswarm/k8smetadata and reference them.

TheoBrigitte:
> The presence of the `--suspend` flag IMO suggests that the command can now be used simply to turn Flux suspension on or off. And that would be great.
> 
> ```
> kubectl-gs update app --namespace ns --name foobar --suspend=true
> kubectl-gs update app --namespace ns --name foobar --suspend=false
> ```

The current implementation does allow for this. Plus `--suspend` is the shorter version for `--suspend=true` which the default behavior of cobra flags.

> The `--version` flag is required, which doesn't make sense with the use case. Also, if I understand the code correctly, the execution of the patch function relies on the condition `len(options.Version) > 0`. I think this should be changed to leave the version untouched.

I agree on this, and will change that.

> Regarding the naming `--suspend`, I think this is a bit too general. I know, someone would love to type fewer letters, but that's what aliases and shells scripts are good for. How about `--suspend-reconciliation`.

This make sense, I will change that even though I am not a fan of long arguments, but I also not feel like using `-s` as alias here (we can still add this later on if needed)

> What happens if the App CR is not reconciled by Flux (not a GitOps-managed resource) and the user uses the command with a version and `--suspend=true`? I think ideally the user should be at least informed in that case.

In that case the label `kustomize.toolkit.fluxcd.io/reconcile=disabled` would be added to the App CR but I would not have any effect. I can adapt this behavior to detect if the app is managed by flux by checking the presence of the `kustomize.toolkit.fluxcd.io/name` label.

pkg/data/domain/app/service.go
TheoBrigitte: Good point

TheoBrigitte:
Changes I made:

- renamed `--suspend` flag to `--suspend-reconciliation`
- move Flux label and annotation to https://github.com/giantswarm/k8smetadata/pull/107
- change logic to only update version when `--version` flag contains a value
- update final output message to reflect change being made

Output examples:
```
$ k gs update app --name prometheus-meta-operator --namespace giantswarm --version 4.75.1-eb3c621412224384484b9d2f6f9cb952040987a5 --suspend-reconciliation=true
App "prometheus-meta-operator" in namespace "giantswarm" updated with version=4.75.1-eb3c621412224384484b9d2f6f9cb952040987a5 added annotations["kustomize.toolkit.fluxcd.io/reconcile"]=disabled
```
```
$ k gs update app --name prometheus-meta-operator --namespace giantswarm --version 4.75.1-eb3c621412224384484b9d2f6f9cb952040987a5 --suspend-reconciliation=false
App "prometheus-meta-operator" in namespace "giantswarm" updated with version=4.75.1-eb3c621412224384484b9d2f6f9cb952040987a5 removed annotations["kustomize.toolkit.fluxcd.io/reconcile"]
```
```
$ k gs update app --name prometheus-meta-operator --namespace giantswarm --version 4.75.1-eb3c621412224384484b9d2f6f9cb952040987a5
App "prometheus-meta-operator" in namespace "giantswarm" updated with version=4.75.1-eb3c621412224384484b9d2f6f9cb952040987a5
```

marians (approved):
Looks good to me! Thanks for the changes. BTW, not looking into code now.

PR #1335, opened 2024-05-31
Add interactive mode to template app command

TheoBrigitte:
### What does this PR do?

This PR adds an interactive mode to the `template app` command. This mode allows users to create a new app interactively by selecting a catalog and a catalog entry from a list of available options.

Users can now create a new app interactively by running `kubectl gs template app --cluster-name myCluster --interactive` and following the prompts.

Review the code and test the new interactive mode.

Yes online doc needs to be updated at https://docs.giantswarm.io/vintage/use-the-api/kubectl-gs/template-app/

No, default bevahior is unchanged.


marians:
This is an interesting experiment, and since some interactive mode or wizard mode in the CLI has been discussed here and there, it's nice to actually try out an example.

I wonder: what motivated you to work on this? What problem are you trying to solve?

As a general notice: I think it makes sense to invest effort into an interactive mode only if we know we want to pursue this for all or most commands.

When testing the implementation here, I think it would require more work to make me want to use it. Here are some things I don't like in particular:

- I should not be required to set a cluster name via the command line flag. Instead, if not given, the interactive mode should allow me to select a cluster.

- There should be some guidance, like what information am I supposed to give in this step? E. g. "Select a catalog to install  an app from".

- The screen layout is very unexpected to me, everything appears displaced. Interactive mode takes the entire terminal window, and the input prompt is in the bottom. Details on the selected catalog appear in the top right (which can be very far away, depending on the window size).

- Loading catalog entries seems to take quite some time (several seconds). I'd like to see some feedback while it happens.

My guess is that it would require a good deal more work to make this a nice experience, and given the limited resources we have, I wouldn't want to give this task to any of our teams currently.

TheoBrigitte:
> I wonder: what motivated you to work on this? What problem are you trying to solve?

I try as much as possible to avoid browser and clicking around type of interactions. I wanted to be able to install an app onto a cluster in a similar way happa would offer.

> * I should not be required to set a cluster name via the command line flag. Instead, if not given, the interactive mode should allow me to select a cluster.

I agree and will change this.

> * There should be some guidance, like what information am I supposed to give in this step? E. g. "Select a catalog to install  an app from".

Will see how feasible is that

> * The screen layout is very unexpected to me, everything appears displaced. Interactive mode takes the entire terminal window, and the input prompt is in the bottom. Details on the selected catalog appear in the top right (which can be very far away, depending on the window size).

That's a drawback from the library I used here. Will see if there's a way to improve this.

> * Loading catalog entries seems to take quite some time (several seconds). I'd like to see some feedback while it happens.

I will add some feedback

pipo02mix:
@TheoBrigitte what we do here?

PR #1352, opened 2024-06-20
Use new releases for CAPA cluster templates

nprokopic:
### What does this PR do?

Towards https://github.com/giantswarm/roadmap/issues/3466

- CAPA only change for new releases: render release version in config instead of cluster-aws version in App resource.

Users must specify `--release` flag when rendering templates for new CAPA clusters.

### What does it look like?

See `--release` flag and `.global.release.version` config set, while App `.spec.version` is not set.

```
kubectl gs template cluster --provider capa --name nikola --organization giantswarm --release 25.0.0
---
apiVersion: v1
data:
  values: |
    global:
      connectivity:
        availabilityZoneUsageLimit: 3
        network: {}
        topology: {}
      controlPlane: {}
      metadata:
        name: nikola
        organization: giantswarm
        preventDeletion: false
      nodePools:
        nodepool0:
          instanceType: m5.xlarge
          maxSize: 10
          minSize: 3
          rootVolumeSizeGB: 8
      providerSpecific: {}
      release:
        version: 25.0.0
kind: ConfigMap
metadata:
  creationTimestamp: null
  labels:
    giantswarm.io/cluster: nikola
  name: nikola-userconfig
  namespace: org-giantswarm
---
kind: App
metadata:
  labels:
  name: nikola
  namespace: org-giantswarm
spec:
  catalog: cluster
  config:
    configMap:
      name: ""
      namespace: ""
    secret:
      name: ""
      namespace: ""
  kubeConfig:
    context:
      name: ""
    inCluster: true
    secret:
      name: ""
      namespace: ""
  name: cluster-aws
  namespace: org-giantswarm
  userConfig:
    configMap:
      name: nikola-userconfig
      namespace: org-giantswarm
  version: ""
```

- https://github.com/giantswarm/roadmap/issues/3466
- https://github.com/giantswarm/rfc/pull/96

Check if I missed to change something.

I don't think so.

Maybe, as `--release` flag is needed for CAPA clusters now, but's a change for CAPA clusters, and I don't think it's a change for kubectl-gs.


nprokopic:
Will fix the tests.

internal/key/key.go
```diff
@@ -100,6 +100,11 @@ func IsPureCAPIProvider(provider string) bool {
 	return contains(PureCAPIProviders(), provider)
 }
 
+// IsCAPIProviderThatUsesReleases returns whether a given provider is a CAPI provider that uses new releases.
+func IsCAPIProviderThatUsesReleases(provider string) bool {
```
njuettner, line 104: what about renaming this function to `IsProviderUsingReleases`

njuettner: it's a bit easier to read IMHO

njuettner: or IsCAPIProviderUsingReleases

internal/key/provider.go
```diff
@@ -41,3 +41,10 @@ func PureCAPIProviders() []string {
 		ProviderCloudDirector,
 	}
 }
+
+// CAPIProvidersThatUseReleases is the list of CAPI providers which are using Release resources.
+func CAPIProvidersThatUseReleases() []string {
```
njuettner, line 46: same here maybe `ProviderUsingReleases` or `CAPIProviderUsingReleases`

njuettner (approved):

internal/key/provider.go
nprokopic: yup, good point, will change these

internal/key/key.go
nprokopic: ✅ renamed it

internal/key/provider.go
nprokopic: ✅ renamed it

PR #1382, opened 2024-07-22
Helpful error messages for invalid subnet split parameters of CAPA private clusters

AndiDog:
### What does this PR do?

The diff looks horrible, sorry. To use the flag `--private-subnet-size` in an error message, I had to make `FlagAWSPrivateSubnetMask` available and because Go doesn't like import cycles (parent uses child items and vice versa), I moved several things into its own `cmd/template/cluster/common` package.

This fix replaces the cidr library's error

```
Error: Num must the power of 2
```

by an actionable error in case the provided subnet size argument is wrong.

### What does it look like?

See code change – two new error cases

My snippet for private clusters didn't work anymore and the error was useless.

It's only for private clusters which right now aren't used much, so I'd say _not now_.

No


cmd/template/cluster/provider/capa.go
```diff
@@ -98,15 +100,18 @@ func templateClusterCAPA(ctx context.Context, k8sClient k8sclient.Interface, out
 		}
 
 		if config.AWS.NetworkVPCCIDR != "" {
-			c, _ := cidr.Parse(config.AWS.NetworkVPCCIDR)
```
AndiDog, line 101: The relevant changes begin here

fiunchinho (approved):

AndiDog:
Test code still needed fixes

PR #1399, opened 2024-08-12
Print release version in `get cluster`

mcharriere:
### What does this PR do?

Print release version in `get cluster` command taken from Cluster CR.

Show correct information about cluster versions.

### What does it look like?
before:
```
$ k gs get clusters -A
NAMESPACE        NAME      AGE    CONDITION   CLUSTER VERSION   PREINSTALLED APPS VERSION   SERVICE PRIORITY   ORGANIZATION   DESCRIPTION
org-giantswarm   golem     651d   READY                                                     highest            giantswarm     golem MC
org-mati         mati225   69m    READY                                                     highest            mati           n/a
```

after:
```
$ ./kubectl-gs get clusters -A
NAMESPACE        NAME      AGE    CONDITION   RELEASE   SERVICE PRIORITY   ORGANIZATION   DESCRIPTION
org-giantswarm   golem     651d   READY       25.0.0    highest            giantswarm     golem MC
org-mati         mati225   69m    READY       29.0.0    highest            mati           n/a
```

none

This changes the old behavior where the default provider was assumed to be CAPO. That's no longer the case, but CAPO is not supported anymore. If there was the need to introduce this provider again, extensive work will be needed to adapt it to the release framework.


marians (approved):
Nice! Please also update https://docs.giantswarm.io/vintage/use-the-api/kubectl-gs/get-clusters/

PR #1403, opened 2024-08-13
Adapt `gitops add ...` commands to CAPI release structure (Breaking)

marians:
### What does this PR do?

The `gitops add workload-cluster>` commands are adapted to work with the new release concept. Previously, they required setting the cluster-app version and the defaults-apps version, and allowed for passing additional default apps configuration. Now setting the release version is required instead.

- The flag `--default-apps-release` got removed
- The flag `--cluster-release` got removed
- The flag `--release` has been added to specify the [release](https://github.com/giantswarm/releases) version to use

https://github.com/giantswarm/roadmap/issues/3630

Yes, PR is here: https://github.com/giantswarm/docs/pull/2278

Yes


mproffitt:
I have tested this on golem and the change works well.

ljakimczuk (approved):

PR #1453, opened 2024-10-09
change link to vintage, since redirect is pointing to a subpage

theBNT:
### What does this PR do?

I noticed that the current link redirects to `https://docs.giantswarm.io/vintage/use-the-api/kubectl-gs/update-cluster/` which is a subpage of the `kubectl gs` docs so i changed the link to point to the vintage URL like the other doc URLs in the README.

They have one less click and are maybe not confused

Decide whether all links should be changed to the vintage links or if the redirects need to be checked/regenerated.

The actual docs not, maybe the redirects from old links to vintage links.

no

no

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,7 +25,7 @@ Check the [installation docs](https://docs.giantswarm.io/vintage/use-the-api/kub
 
 ## Documentation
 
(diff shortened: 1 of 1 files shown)


marians (approved):
Thank you!

PR #1471, opened 2024-10-31
Update telemetrydeck to fix unique user count problem

marians:
### What does this PR do?

The same user hash got submitted for all users. This was fixed in https://github.com/giantswarm/telemetrydeck-go/pull/10.

Also this PR removes debug logging related to telemetry, which was a leftover.

Users should definitely no see any debugging log messages regarding telemetry.

The main change is invisible to users.

No


marians:
Another user got `cf4d9c1d83f`

uvegla (approved):

PR #1489, opened 2024-11-15
Adjust columns for kubectl gs get releases

njuettner:
### What does this PR do?

Before:

```
VERSION                 STATUS   AGE   KUBERNETES   CONTAINER LINUX   COREDNS   CALICO
cloud-director-27.0.0   ACTIVE   44h   1.27.16      n/a                         n/a
```

After:

```
VERSION                 STATUS   AGE   KUBERNETES   FLATCAR    CILIUM   COREDNS   OBSERVABILITY BUNDLE   SECURITY BUNDLE
cloud-director-27.0.0   ACTIVE   44h   1.27.16      3815.2.5   0.25.1   1.21.0    1.5.3                  1.8.0
```

### What does it look like?

- [ ] CHANGELOG.md has been updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ## [Unreleased]
 
+### Changed
+
+- Adjust columns for `kubectl gs get releases`.
+
 ## [4.4.0] - 2024-11-13
 
 ### Added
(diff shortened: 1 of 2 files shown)


AverageMarcus:
I don't think we ever refer to it as "container linux", at least not that I've seen. We always say "Flatcar" in our releases and when talking to customers so it should help with consistency.

AverageMarcus (approved):

marians:
All right

Gacko:
Out of curiosity, maybe there's an UX reason for it: Why don't we use the CRD native possibility of displaying fields in just a normal `kubectl get`? As far as I can tell we're already doing this for some.

PR #1500, opened 2024-11-26
Update github.com/ProtonMail/gopenpgp to v3, adapt to API change

marians:
### What does this PR do?

Bumps `github.com/ProtonMail/gopenpgp` to v3.

The intend is that there is no effect

Testing. The function is used in these commands:

- `gitops add encryption`
- `gitops add management-cluster`

No

No

No


uvegla (approved):

uvegla:
LGTM, tested locally.

PR #1624, opened 2025-04-01
Add in-cluster flag to add the required kubeConfig property

pipo02mix:
### What does this PR do?
Add `kubeConfig` property to GitOps adds command to get a functional app CR
(Please set a descriptive PR title. Use this space for additional explanations.)

### What does it look like?


cmd/gitops/add/app/runner.go
```diff
@@ -68,6 +68,14 @@ func (r *runner) run(ctx context.Context, cmd *cobra.Command, args []string) err
 		config.AppName = config.App
 	}
 
+	if cmd.Flags().Changed("inCluster") && r.flag.InCluster {
```
pipo02mix, line 71: Only add it when the flag is explicitly passed. This ensure the behavior has not change by default

mproffitt (approved):

pipo02mix:
Now default behavior is the same, you need to explicitly pass the flag to get the kubeConfig property

PR #1931, opened 2026-04-13
Make `deploy chart` more flexible with `--update-only` and `--namespace`

AndiDog:
I tried to use this to edit `HelmRelease`/`OCIRepository` for MC operators. The first issue is that the command upserts, but we don't want to _create_ objects, so I introduced `--update-only` as safety measure. Second, `--organization` was the only way to choose the object namespace, while our operators are in namespace `giantswarm`, so I added `--namespace` as alternative. The last issue, not solved in this PR, is prefixing the object names with `<cluster>-` as expected for workload clusters.

Essentially, I'm not sure if MC operators _should_ be deployable with this tool, so this PR is still a draft to show the issues that arose for Giant Swarm developers. Let's decide if we want to have that functionality in this tool, as we did before with `kubectl gs update app`. Once we've done that, there's another issue: suspending Flux reconciliation is needed for our purposes.

No

### DIFF
--- a/cmd/deploy/chart/flag.go
+++ b/cmd/deploy/chart/flag.go
@@ -2,15 +2,11 @@ package chart
 
 import (
 	"fmt"
-	"regexp"
 	"strings"
 
 	"github.com/giantswarm/microerror"
 	"github.com/spf13/cobra"
 )
-
-// semverRe matches a basic semver string like "1.2.3" or "v1.2.3".
-var semverRe = regexp.MustCompile(`^v?\d+\.\d+\.\d+$`)
 
 const (
 	flagChartName         = "chart-name"
@@ -95,10 +91,6 @@ func (f *flag) Validate() error {
 	}
 	if f.TargetNS == "" {
 		return microerror.Maskf(invalidFlagError, "--%s must not be empty", flagTargetNS)
-	}
-
-	if f.Version != "" && !semverRe.MatchString(f.Version) {
-		return microerror.Maskf(invalidFlagError, "--%s must be a valid semver version (e.g. 1.2.3), got %q", flagVersion, f.Version)
 	}
 
 	if f.AutoUpgrade != "" {
--- a/cmd/deploy/chart/flag.go
+++ b/cmd/deploy/chart/flag.go
@@ -11,6 +11,7 @@ import (
 const (
 	flagChartName         = "chart-name"
 	flagOrganization      = "organization"
+	flagNamespace         = "namespace"
 	flagCluster           = "target-cluster"
 	flagTargetNS          = "target-namespace"
 	flagOCIURLPrefix      = "oci-url-prefix"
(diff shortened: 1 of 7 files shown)


cmd/deploy/chart/flag.go
```diff
@@ -2,19 +2,16 @@ package chart
...
-	"regexp"
 	"strings"
 
 	"github.com/giantswarm/microerror"
 	"github.com/spf13/cobra"
 )
 
-// semverRe matches a basic semver string like "1.2.3" or "v1.2.3".
-var semverRe = regexp.MustCompile(`^v?\d+\.\d+\.\d+$`)
-
```
marians, line 14: Why remove this?

AndiDog: This was too restrictive. It didn't allow `X.Y.Z-COMMIT`.

AndiDog:
Closing since I won't work on this anymore. I'm not successfully using a very simple script that labels OCIRepository with `kustomize.toolkit.fluxcd.io/reconcile=disabled` and triggers reconciliation (or for unsuspend: unlabel+trigger).

marians:
@AndiDog Is that use case (suspending and overwriting an existing deployment) maybe something for devctl?

AndiDog:
> @AndiDog Is that use case (suspending and overwriting an existing deployment) maybe something for devctl?

@marians I'm not sure if it's `dev` or `ops`-ctl 😆... We had the functionality in `kubectl gs update app` previously, meaning it could be used outside Giant Swarm. The scripts I'm using for now (see below) could easily be integrated into such a command, I guess. It doesn't seem to fit `deploy chart`, though, since that already behaves differently (create-or-update, which to me is surprising) than `update app` did.

---

```
gs_update_operator() {
	(
		set -euo pipefail
		[ $# = 2 ] || { >&2 echo "Usage: $0 capa-iam-operator X.Y.Z-COMMIT # for example"; exit 1; }

		name="${1}"
		version="${2}"

		(set -x; kubectl label -n giantswarm ocirepository.source.toolkit.fluxcd.io "${name}" --overwrite kustomize.toolkit.fluxcd.io/reconcile=disabled)

		(set -x; kubectl patch -n giantswarm ocirepository.source.toolkit.fluxcd.io "${name}" --type=merge -p "{\"spec\": {\"ref\": {\"semver\": \"${version}\"}}}")

		(set -x; flux reconcile source oci -n giantswarm "${name}")
		(set -x; flux reconcile hr -n giantswarm "${name}")
	)
}

# Pendant for `gs_update_operator`
gs_rollback_operator() {
	(
		set -euo pipefail
		[ $# = 1 ] || { >&2 echo "Usage: $0 capa-iam-operator # for example"; exit 1; }

		name="${1}"

		(set -x; kubectl label -n giantswarm ocirepository.source.toolkit.fluxcd.io "${name}" --overwrite kustomize.toolkit.fluxcd.io/reconcile-)
		(set -x; flux reconcile kustomization -n flux-giantswarm collections)
		(set -x; flux reconcile source oci -n giantswarm "${name}")
		(set -x; flux reconcile hr -n giantswarm "${name}")
	)
}
```

PR #1933, opened 2026-04-14
Remove unused code

AndiDog:
Everything that Claude Code could find as dead 😉


fhielpos (approved):
I guess we rarely assert these kind of errors in code so LGTM.

PR #1942, opened 2026-04-28
Add structured auth to login command

pipo02mix:
Towards https://github.com/giantswarm/adidas/issues/1414

 Added new flags to support structured auth (optional, as it will try to discover automatically):
  - --issuer — OIDC issuer URL override.
  - --client-id — OIDC client ID override.
  - --ca-file — path to WC API server CA, bypasses the cluster-values ConfigMap fetch.

Implemented a new OIDC WC flow
  - detectStructuredAuth checks if structured auth is configured and parses auth-config.yaml (handles base64 encoding) from the KubeadmControlPlane to extract issuer URL + audience.
  - fetchClusterCA reads the WC CA either from --ca-file or from <cluster>-cluster-values ConfigMap.
  - createOIDCKubeconfig runs the browser-based OIDC flow (handleDirectOIDC) directly against the issuer (e.g. Azure AD), uses RedirectURI: "/" for compatibility with apps registered as plain http://localhost:<port>, verifies the token against the WC API, writes a kubeconfig context named gs-<mc>-<wc>-oidc, and caches tokens via
  credentialcache.WriteWithLock.
  - handleDirectDeviceOIDC returns a clear error since device-auth isn't supported here.


cmd/login/command.go
```diff
@@ -76,6 +75,24 @@ Workload cluster client certificate:
...
     --` + flagWCCertGroups + ` admins \
     --` + flagWCCertTTL + ` 3h
+
+Workload cluster direct OIDC (Kubernetes structured authentication),
+auto-discovering issuer URL, client ID and CA from the management cluster:
+
+  kubectl gs login mymc \
+    --` + flagWCName + ` gir0y \
+    --` + flagWCOrganization + ` acme \
+    --` + flagStructuredAuth + `
```
pipo02mix: I am not sure about the naming, open to discussion. Now the command
```
/kubectl-gs login graveler \
      --workload-cluster fer-test-auth \
      --organization fer \
      --structured-auth
```
Take the client id, issuer and CA from MC. If you pass the flags it will not need MC connection. Is that clear enough?

stone-z:
I haven't had the time yet to look into this at a technical level, but wanted to say early that I would prefer if kgs detected the correct type of login to use for a cluster and then just did that without the user needing to know about it. Just spitballing: maybe this means we wire more information through athena and use that to inform kgs.

pipo02mix:
> I haven't had the time yet to look into this at a technical level, but wanted to say early that I would prefer if kgs detected the correct type of login to use for a cluster and then just did that without the user needing to know about it. Just spitballing: maybe this means we wire more information through athena and use that to inform kgs.

But now there are 3 behaviors (before my change)
- Login MC with OIDC
- Login WC with OIDC (only for customers having Athena/Dex in the WC)
- Login WC with certs

I am adding
- Login WC with OIDC using structured auth (without the need for Athena/Dex)

How does kgs login know which one to use? I am happy to think more, but we should not break backward compatibility.

marians:
This PR is somehow adding a new authentication mode for workload clusters (WC). We had a mode for management cluster (MC) already and three for WC:

1. MC
2. WC OIDC via Dex
3. WC client cert
4. WC AWS IAM

The new one:

5. WC direct OIDC

What seems to be special in the new mode is auto-discovery of OIDC provider details (which can be short-cut with `--issuer`, `--client-id`, and `--ca-file`).

The new flag `--structured-auth` to enable the new 5th mode has unfortunate naming. Structured Authentication should be a cluster admin's concern, not the user's. Ideally, users don't have to know how the cluster is configured in order to log in.

Based on the usage, the command is supposed to be called like this:

```nohighlight
kubectl gs login <installation> \
    --workload-cluster <cluster-name> \
    --organization <org-name> \
    --structured-auth
```

What if the `--structured-auth` flag didn't exist? Could we simply let the command do the discovery and adapt to the cluster setup, using the appropriate technique?

In that case I would rename the three new flags to `--oidc-issuer`, `--oidc-client-id`, and `--api-ca-file` to make them more specific.

---

There are some more UX concerns with this PR:

1. `--connector-id` is overloaded. Its description was rewritten to: "Dex connector to use… or OIDC client-id for selecting among multiple structured authentication issuers on a workload cluster." That makes one flag mean two unrelated things, and in mode 5 it duplicates `--client-id` exactly. Both match against `iss.ClientID` in `structuredauth.go:101 / :118.` Pick one. I'd keep `--connector-id` for Dex only, and use `--client-id` (or `--oidc-client-id`) to disambiguate issuers.

2. Silent behavior change for no-arg invocation. runner.go:67 now shows help when no arg and no --workload-cluster/context is given. Previously this tried to reuse the current context (and could error with selectedContextNonCompatibleError, see the test diff). This is a severe behavior change that I know some customers (who start their session just calling `kubectl gs login` -- just in case) won't like.

3. Help text and error messages leak implementation. The example block says "no MC lookup of KubeadmControlPlane or cluster-values ConfigMap is performed", and the not-found error tells the user "ensure the KubeadmControlPlane has an auth-config file". A user logging into their cluster shouldn't need to know what a KCP is. Reframe in user terms (e.g. "could not auto-detect the OIDC issuer for this cluster. Pass --oidc-issuer and --oidc-client-id").

4. The "skip MC validation" message is also implementation-flavored. "Using current kubeconfig context for MC access (skipping MC OIDC validation)" -- the parenthetical is internal noise. Something like "Using the current kubeconfig context to reach the management cluster." is enough.

5. Other examples in the usage info use `mywc` as workload cluster name. Would be easiest to re-use that one.

fhielpos:
I agree with Marian, `structured-auth` shouldn't be of users concerns. Ideally the `k-gs login` command should be as minimal as possible from a user perspective.

pipo02mix:
@fhielpos @marians PTAL

marians:
@pipo02mix Is the PR description up-to-date?

pipo02mix:
@marians now it is

marians (approved):
LGTM. Thanks!

pipo02mix:
I have tested with old auth-bundle WC installations and EKS, and it works. @fhielpos WDYT? can I merge?

PR #1971, opened 2026-05-19
Enable PKCE on the direct OIDC flow

pipo02mix:
Public OIDC clients (e.g. Okta SPA/Native apps) registered without a `client_secret` rejected the token exchange with "invalid_client" . Implement PCKE to allow login using our login.


puja108 (approved):
LGTM, didn't test

PR #1972, opened 2026-05-19
Prompt user to pick OIDC issuer when multiple are configured

pipo02mix:
When a workload cluster's structured-auth config exposes more than one JWT issuer and no `--oidc-client-id / --oidc-issuer` flag is set, show a numbered menu and read the selection from stdin instead of erroring out. Falls back to the prior error (with the available-issuers list) when stdin is not a TTY, so scripted callers stay informative.


puja108 (commented):
minor nits from Claude:

Prompt goes to stderr, but rest of wc.go writes to r.stdout. The interactive prompt is written to r.stderr ([cmd/login/wc.go:202](https://github.com/giantswarm/kubectl-gs/blob/claude/review-pr-1972-WKzJV/cmd/login/wc.go#L202)), while every other user-facing message in createClusterKubeconfig uses r.stdout (wc.go:138-159). Writing prompts to stderr is defensible (keeps stdout pipe-clean), but it's inconsistent here. Worth a comment explaining the choice — or just write to r.stdout to match surrounding code.

TTY check only inspects stdin, not the writer. If a user runs kubectl gs login ... 2>/dev/null with stdin still attached to a TTY, the prompt will be silently swallowed and the command will appear to hang. Consider also checking term.IsTerminal(int(os.Stderr.Fd())) (or whichever stream the prompt is written to).

Single failed entry kills the session. A typo like "2 " succeeds (trimmed), but "two" returns an error and the whole login aborts. A re-prompt loop (with a small attempt cap, e.g. 3) would be more forgiving and matches how most CLIs handle interactive selection. Not blocking, but a UX win.

otherwise LGTM

puja108 (approved):

PR #1977, opened 2026-05-20
feat(CAPA): derive the number of VPC chunks from `--az-usage-limit` to support smaller CIDRs

iuriaranda:
### What does this PR do?

`template cluster --provider=capa`: derive the number of VPC chunks from `--az-usage-limit` instead of always splitting the VPC into 4 equal parts. This enables small single-AZ VPC layouts (e.g. a `/24` VPC with a `/25` private subnet and a `/26` public subnet, previously rejected with "you must specify a smaller private subnet size"). Existing 2- and 3-AZ layouts are unchanged. Single-AZ layouts now use the lower half of the VPC for public subnets and the upper half for private subnets, so the rendered private subnet address may shift compared to previous releases (the subnet size and count are unchanged). The clearer error message for VPC/AZ/subnet-size combinations that cannot be satisfied also replaces a confusing per-chunk message.

Towards https://github.com/giantswarm/giantswarm/issues/36485


cmd/template/cluster/provider/capa.go
AndiDog:
```suggestion
			vpcCidr, err := cidr.Parse(config.AWS.NetworkVPCCIDR)
```

AndiDog: Not sure if the PR suggestion feature works – the diff doesn't render as intended to me. I suggest so rename `c` to `vpcCidr` for readability.

AndiDog (approved):

cmd/template/cluster/provider/capa.go
iuriaranda: done

PR #1979, opened 2026-05-21
fix: Tolerate OIDC providers that do not issue a refresh token

pipo02mix:
Coming from https://gigantic.slack.com/archives/C092KTAL79V/p1779332362504319?thread_ts=1778715187.880409&cid=C092KTAL79V


fhielpos (approved):
LGTM, changelog entry is maybe a bit long but it's fine if you want to keep it that way

PR #2058, opened 2026-07-16
Avoid MC connection in login command

pipo02mix:
### What does this PR do?

Towards https://github.com/giantswarm/giantswarm/issues/37183. Makes optional the MC connection in OIDC login flows

### What does it look like?

An optional flag added; it does not affect existing command behavior

done in this [PR](https://github.com/giantswarm/docs/pull/3283)

### DIFF
--- a/cmd/login/wc.go
+++ b/cmd/login/wc.go
@@ -3,6 +3,7 @@ package login
 import (
 	"context"
 	"fmt"
+	"net/url"
 	"os"
 	"regexp"
 	"strings"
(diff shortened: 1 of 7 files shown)


cmd/login/wc.go
```diff
@@ -454,3 +491,18 @@ func isEKS(c *unstructured.Unstructured) bool {
...
+// normalizeAPIEndpoint validates the user-supplied workload cluster API server
+// endpoint and returns it with an https:// scheme prepended when absent.
+func normalizeAPIEndpoint(endpoint string) (string, error) {
+	if !strings.Contains(endpoint, "://") {
+		endpoint = "https://" + endpoint
+	}
+
+	u, err := url.Parse(endpoint)
+	if err != nil || u.Host == "" {
+		return "", microerror.Maskf(invalidFlagError, "--%s %q is not a valid URL", flagWCAPIEndpoint, endpoint)
```
stone-z, line 550: Not sure if this covers everything; intuitively, I think there are probably some weird cases that would slip through here and shouldn't. But I don't have a concrete case, so I'll just ask:
Currently, this would accept an http endpoint - any reason not to require https, since this is for auth?

stone-z (approved):
One comment on the URI validation, but this seems reasonable to me

cmd/login/wc.go
pipo02mix: I was thinking of the internal API case where the endpoint is not exposed via SSL. Let me double-check

pipo02mix: Added check below

stone-z (approved):
Thanks Fer!

PR #2130, opened 2026-09-10
fix: render --service-priority as global.metadata.servicePriority

ljakimczuk:
### What does this PR do?

Makes `template cluster --service-priority` actually reach the rendered manifests. The flag
was registered and validated, and its value was copied into `common.ClusterConfig`, but
nothing read it back out — so it was a silent no-op for all six providers.

The value is now rendered as `global.metadata.servicePriority`, mirroring how
`--prevent-deletion` reaches `global.metadata.preventDeletion`: one field added to each
provider's `Metadata` struct in `templates/*/types.go`, and one pass-through line in each
`Build*ClusterConfig`. 12 one-line changes, no new logic.

`--service-priority medium` and `--service-priority lowest` now do what they say. Before
this, they were silently ignored and every templated cluster came out as `highest`.

**Clusters templated without the flag are unaffected.** The flag defaults to `highest` and
`global.metadata.servicePriority` defaults to `highest` in the chart, so rendered output is
byte-identical unless a non-default value is passed. Existing golden files are unchanged.

### What does it look like?

    kubectl gs template cluster --provider capa --name test --organization test \
      --release 35.0.0 --service-priority medium

The user ConfigMap values gain one key:

```yaml
global:
  metadata:
    name: test
    organization: test
    servicePriority: medium
```

which the `cluster` library chart turns into `giantswarm.io/service-priority: medium` on the
cluster resources (`helm/cluster/templates/_helpers.tpl`, `cluster.labels.common`), and which
then shows up in the `SERVICE PRIORITY` column of `kubectl gs get clusters`.

`--service-priority` was added in #827 and given its `highest` default in #812, but it was
only ever wired into the **vintage** AWS/Azure templating paths, which set the label directly
on a typed `Cluster` CR. Those blocks were deleted along with the vintage providers in #1871
(02776da7), shipping in v5.0.0, and the flag, its validation and the `ClusterConfig` field
were left behind with no consumer.

The same commit also broke `--label`, which was fixed the same day in #1877 and shipped in
v5.0.1. `--service-priority` was missed in that fix, so this is the remaining half of that
regression rather than an intentional deprecation — nothing in `CHANGELOG.md` announced
dropping the flag.

CAPI-side the label is a first-class value, not a generic label:
`global.metadata.servicePriority` is schema-**required** with `enum: [highest, medium, lowest]`
and `default: highest`, present in all six provider charts (`cluster-aws`, `-azure`, `-aks`,
`-eks`, `-vsphere`, `-cloud-director`), per the
[classify-cluster-priority RFC](https://github.com/giantswarm/rfc/tree/main/classify-cluster-priority).

Two things:

1. **Sanity-check the mechanism.** Setting the dedicated `servicePriority` value is, I believe,
   the right route. Going via `global.metadata.labels` instead would collide with the chart's
   own emission of `giantswarm.io/service-priority` in `cluster.labels.common` and produce a
   duplicate YAML key, so I deliberately did not touch the labels map.
2. **A follow-up worth its own issue:** `pkg/labels/labels.go` still exempts
   `giantswarm.io/service-priority` from the forbidden-`giantswarm.io`-prefix check. That
   exemption made sense in the vintage code, which read the key back out of `config.Labels`,
   but today `--label giantswarm.io/service-priority=...` walks straight into the duplicate-key
   collision above. Happy to open that separately rather than widen this PR.

No golden test case is included — the change is pure pass-through and the existing goldens are
unaffected. If you'd like one pinned, add `ServicePriority: "medium"` to a `Test_run` case and
re

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,8 +12,9 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 - `template cluster`: `--service-priority` takes effect again. The flag was accepted and validated
   but its value never reached the rendered manifests, because its only consumers were the vintage
   AWS/Azure providers removed in v5.0.0. It is now rendered as `global.metadata.servicePriority`
-  for every provider. Clusters templated without the flag are unaffected, since the flag default
-  and the chart default are both `highest`.
+  for every provider. Because the flag defaults to `highest`, templated manifests now always
+  carry `global.metadata.servicePriority`; this is semantically a no-op against the chart
+  default of `highest`, but re-templated clusters will show the added key in a GitOps diff.
 
 - Usage telemetry: the version is also sent as `TelemetryDeck.AppInfo.version`, the parameter the TelemetryDeck dashboard's standard "App Versions" insight reads (it was only in the payload key `appVersion`, so that chart stayed empty). telemetrydeck-go v0.2.0.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project's packages adheres to [Semantic Versioning](http://semver.org/s
 
 ### Fixed
 
+- `template cluster`: `--service-priority` takes effect again. The flag was accepted and validated
+  but its value never reached the rendered manifests, because its only consumers were the vintage
+  AWS/Azure providers removed in v5.0.0. It is now rendered as `global.metadata.servicePriority`
+  for every provider. Clusters templated without the flag are unaffected, since the flag default
+  and the chart default are both `highest`.
+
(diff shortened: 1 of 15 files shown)


uvegla (approved):

mproffitt (approved):
Auto approved by claude via `pr-triage`

During review the following nits were noticed but do not block acceptance
- PR text says output is "byte-identical" without the flag; it isn't — the ConfigMap now always gains `servicePriority: highest` because the flag defaults to `highest`, at cmd/template/cluster/flags/flag.go:279. Semantically a no-op against the chart default, but GitOps diffs on re-templated clusters will show the new key.
- No test pins the new pass-through. This flag has now been silently dropped once already; one golden case with `ServicePriority: "medium"` would stop it happening a third time — cmd/template/cluster/provider/capa.go:353.

